Issue #1, opened 2015-09-09
terminate called after throwing an instance of 'std::invalid_argument'


frodrigo:
I got en error ruining your french_towns.req test set.
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoul
Abandon (core dumped)
```

```
vroom -a 127.0.0.1 -p 5000 -i french_towns.req
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoul

Program received signal SIGABRT, Aborted.
0x00007ffff71d45f8 in raise () from /usr/lib/libc.so.6
(gdb) where
#0  0x00007ffff71d45f8 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff71d5a7a in abort () from /usr/lib/libc.so.6
#2  0x00007ffff7ae8b3d in __gnu_cxx::__verbose_terminate_handler ()
    at /build/gcc-multilib/src/gcc-5.2.0/libstdc++-v3/libsupc++/vterminate.cc:95
#3  0x00007ffff7ae6996 in __cxxabiv1::__terminate (handler=<optimized out>)
    at /build/gcc-multilib/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:47
#4  0x00007ffff7ae69e1 in std::terminate () at /build/gcc-multilib/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_terminate.cc:57
#5  0x00007ffff7ae6bf8 in __cxxabiv1::__cxa_throw (obj=obj@entry=0x6c7ef0, tinfo=0x7ffff7dcdaa8 <typeinfo for std::invalid_argument>, 
    dest=0x7ffff7afc140 <std::invalid_argument::~invalid_argument()>)
    at /build/gcc-multilib/src/gcc-5.2.0/libstdc++-v3/libsupc++/eh_throw.cc:87
#6  0x00007ffff7b0f9df in std::__throw_invalid_argument (__s=0x425626 "stoul")
    at /build/gcc-multilib/src/gcc-5.2.0/libstdc++-v3/src/c++11/functexcept.cc:82
#7  0x0000000000412fa0 in osrm_wrapper::load_matrix(std::vector<std::pair<double, double>, std::allocator<std::pair<double, double> > > const&) ()
#8  0x000000000040f497 in tsp::tsp(cl_args_t const&) ()
#9  0x00000000004081cc in solve_atsp(cl_args_t const&) ()
[...]

PR #3, opened 2015-09-10
Fix socket receive buffer size #1

frodrigo:


jcoupey:
To avoid messing with the master branch, I added your commits to a new branch improve_socket_handling.

PR #11, opened 2015-11-07
Use GET to access OSRM in place of POST

frodrigo:
Use GET to access OSRM in place of POST, since there is nothing to POST and OSRM 4.8.1 return error on POST to /table

### DIFF
--- a/src/loaders/osrm_wrapper.h
+++ b/src/loaders/osrm_wrapper.h
@@ -38,7 +38,7 @@ private:
   std::string build_query(const std::vector<std::pair<double, double>>& locations, 
                           std::string service, std::string extra_args = "") const{
     // Building query for osrm-routed
-    std::string query = "POST /" + service + "?";
+    std::string query = "GET /" + service + "?";
 
     // Adding locations.
     for(auto const& location: locations){


jcoupey:
This was a workaround to avoid long queries being truncated because of length limitations with GET. Related posts for OSRM:
https://github.com/Project-OSRM/osrm-backend/issues/1293
https://github.com/Project-OSRM/osrm-backend/pull/1502

I will check out to the latest 4.8.1 and try to figure out the best way to deal with long queries.

jcoupey:
Seriously thinking about using libosrm now...

frodrigo:
But, using libosrm force using a local OSRM.

jcoupey:
I think POST should be ok with table requests on OSRM 4.8.1. From what I get from 1502, the change that broke things for the current osrm_wrapper code is the injection of a `?` into the request (clearly visible from osrm-routed output).

jcoupey:
>But, using libosrm force using a local OSRM

Yep, this is the only drawback.

jcoupey:
Have you tried using a remote OSRM instance with the current code? I would appreciate some feedback on this as I'm always working on the same machine.

frodrigo:
I'm only test with remote OSRM

frodrigo:
I'm not sure if I understand all the pieces. But a POST request must contains a body. It's not the case with osrm_wrapper, always use the query.

jcoupey:
Yes, implementing POST queries would really require to do things properly in the wrapper. I'll just let things as they are for now as it is not really a limitation for vroom use.

> I'm only test with remote OSRM

Ok thanks, so it is working fine in this setting.

PR #19, opened 2016-01-10
Minor fix to parse response of osrm v4.6.0

giraldeau:
The new request result for distance_table is not contained into a dict.

This change is backward compatible with the previous OSRM version.

Signed-off-by: Francis Giraldeau <francis.giraldeau@gmail.com>

### DIFF
--- a/src/loaders/osrm_wrapper.h
+++ b/src/loaders/osrm_wrapper.h
@@ -135,7 +135,7 @@ public:
     assert(response.find("Bad Request") == std::string::npos);
 
     // Removing headers.
-    std::string distance_key = "{\"distance_table\":[";
+    std::string distance_key = "\"distance_table\":[";
     size_t table_start = response.find(distance_key);
     assert(table_start != std::string::npos);
 


jcoupey:
Thanks for your input. I take it you mean v4.9.0 of OSRM (not 4.6.0), which contains new keys "destination_coordinates" and "source_coordinates".

I didn't try v4.9.0 yet but from what I see in the [table response description](https://github.com/Project-OSRM/osrm-backend/wiki/Server-api#example-2) the trouble would rather be to avoid getting the new keys at the end. Or maybe the "distance_table" is always the last element in the response?

giraldeau:
The json data may have keys in different order, we cannot assume that
destination_coordinates is the first key in the dict. In fact, we should
rather load the reply as a proper json object instead of parsing the string.

Le lun. 11 janv. 2016 à 05:00, Julien Coupey <notifications@github.com> a
écrit :

jcoupey:
I do agree that dealing with proper json objects would be a much cleaner approach. I've been considering adding a dependency to handle json but this is clearly not a priority right now, so it won't happen soon.

> The json data may have keys in different order

Yes of course but in practice I've always found the order in the OSRM outputs consistent enough (so far) to be relied on for such simple purposes.

Your fix is fine if "distance_table" is always the last entry (hence my question).

jcoupey:
Just made a quick check with v4.9.0 and merged your commit into `develop`. Thanks for pointing this out.

PR #22, opened 2016-01-23
Use Euclidian distance in a virtual matrix

frodrigo:
I try to use euclidean distance in place of matrix. Idea is test how big instance vroom can resolve, without having to store a such large matrix in memory.

It's just conceptual, it's not working (not even compiling)


jcoupey:
My guess is that the computing times will become unreasonably high before storing the matrix will actually become a problem but I really didn't investigate this. What amount of locations do you have in mind when you say "large"?

On the other hand, re-computing the euclidean distance each time a distance is required (that's very often) will overload the computing time itself.

frodrigo:
I was thinking about starting from 30 000 up to 200 000 and if still doable up to million.

Nobody live will be saved with this try, nor man put on Mars, it's just for fun.

jcoupey:
Sounds fun to me too. ;-) Besides, pushing the limits is often a great way to spot improvements.

As this may not be limited to matrix impact on memory, I merged your commit in a dedicated branch `experimental_huge_instances` to play around freely without messing up the `develop` branch. Also set-up issue #23 to further discuss this.

Issue #47, opened 2016-09-27
Using custom matrix from JSON-object instead of OSRM.


PattyDePuh:
Hey folks, 
On [my fork](https://github.com/PattyDePuh/vroom/tree/feature-matrix-via-json-loader) i implemented the first iteration of the `json-uploader`, another option to provide Cost-matrix, instead of OSRM. With the json-uploader, you can use the matrix from the JSON object, that is given to the application. Run the application with the `-j` flag to unlock the feature:
Usage example: `./vroom -i /path/to/json.txt -j`
[test8u.json.txt](https://github.com/VROOM-Project/vroom/files/495256/test8u.json.txt)
Output with above file:
`{"routes":[{"steps":[0,2,3,7,6,1,4,5,8],"cost":8362,"vehicle":1}],"solution":{"computing_times":{"loading":0,"solving":0},"cost":8362},"code":0}` 
(VROOM found one of the optimal solutions for the above instance.)
Because there are no location coordinates necessary in this case, the `"jobs"`-key becomes obsolet and the `"vehicles"`-key provides with `"start"` and `"end"` the desired start and end-points associated with the columns from the matrix.
For roundtrips currently you can't simply set `"start" = "end"` because of an assert in one of the solvers; Gonna' look on this one...

jcoupey:
Hey, thanks for sharing this! 

This feature might be a useful addition. My concerns here are:

1. The TSPLIB format support is the way this is currently implemented.  Maybe you have use-cases in mind where it is not possible to use it? So far, this looks really the same as displaying the explicit matrix in a plain TSPLIB file with the additional keywords: `START:  0` and `END: 8`.

2. This breaks the API defined at https://github.com/VROOM-Project/vroom/blob/master/docs/API.md ; I think we would need to keep a `jobs` array in input. Maybe make the locations optional if a matrix is provided, and require that ids actually describe matrix indexes. This is mandatory if we need to add other keys in jobs for future features.
Ideally, parsing a file with a `matrix` key should reuse the same existing code for json parsing (currently in `osrm_loader`) and just add the matrix stuff on top. I realize this is not possible at the moment but should be easier in the future with #44 where I plan to separate the input parsing and the different loaders.

Thanks for your input, I'm happy to discuss this further!

PattyDePuh:
[...]

Issue #44, opened 2016-09-12
Problem and heuristics refactor


jcoupey:
For future evolutions, the code structure should allow to:
- define and expose different types of problems in an extensible way;
- add dedicated solving strategies as "plugins", or somehow bind each strategy to the relevant problem.

The current TSP-solving code should be refactored as the first plugin. This probably means adding classes to describe objects that don't exist as such so far (e.g. solutions, maybe routes, jobs, vehicles...).

PR #53, opened 2016-11-09
Feature matrix via json loader

PattyDePuh:


jcoupey:
Thanks for submitting this !

As I commented in #47, I think we should wait for the work related to #44 to be merged in develop so your addition proposal will better fit in the (new) code base. The logic here is to completely separate `osrm` calls from the parsing and problem defining itself. You can check the current advancement for this on the `refactor-problem-heuristics` branch. Most relevant for this PR are:
- the `utils/input_parser.*` files where all the json input parsing is done.
- the `structures/vroom/*` files defining useful structs, allowing to defined locations with optional coords (with this matrix thing in mind).

Hopefully, this will also make things much clearer with regard to the problem definition and the possible extensions. I'm sorry this is taking quite a while...

PattyDePuh:
Hey Julien,
i know, that you are currently changing the inner architecture
and the pull-request was never meant to leave my fork. 
Was a miss-click from my side, still learning Git ... 
Thats why excuse me for bothering you and take your time.

jcoupey:
Never mind, I feel it's great that your are still playing around with this implementation idea! At least you got a few hints on what I have in mind for the new structure. I will have to describe it at some point anyway so now at least I started.

About branching: I try to keep `master` for releases so PR are expected against `develop`. You can't really figure this out, I'll have to get around writing a `contributing.md` file.

PR #55, opened 2016-12-01
C++11 compilation for android application.

fijemax:
@jcoupey new pull request
I am currently compiling and linking Vroom in an android application. The native c++ android code source is compiled with std=c++11 parameter.

The "auto" type in lambda functions parameters wasn't accepte in c++11, I replace by real type.

With this modification Vroom can be compil as static library for c++11 or c++14 project.


jcoupey:
@fijemax I removed the lambdas used to sort edges lists as it makes more sense to overload `operator<` for edges.

Issue #72, opened 2017-12-06
Jobs and vehicles object handling


jcoupey:
Currently all jobs and vehicles are created within the scope of `input::add_job` and `input::add_vehicle`. This design is a result of the need to initialize the const `index` values (index in the matrix) for all embedded locations.

This is overly restrictive as for further evolutions we might want to create all kind of jobs and vehicles using a range of ctors (rather than having a bunch of different `add_job` and `add_vehicle`).

Also in the light of #42 the typical usage would be: users create relevant jobs and vehicles objects and add them to whatever `input` object they want before solving.

PR #73, opened 2017-12-07
Refactor jobs/vehicles creation and addition.

jcoupey:
## Issue

#72 

## Tasks

 - [x] modify jobs/vehicles ctors + `input::add_job` and `input::add_vehicles`
 - [x] adjust current parsing in `input_parser`
 - [x] update `libvroom.cpp` example
 - [x] review

### DIFF
--- a/examples/libvroom.cpp
+++ b/examples/libvroom.cpp
@@ -8,6 +8,8 @@
 
 #include "../src/routing/routed_wrapper.h"
 #include "../src/structures/vroom/input/input.h"
+#include "../src/structures/vroom/job.h"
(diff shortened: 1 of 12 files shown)


jcoupey:
Implementation notes: all locations can now be constructed with or without providing the corresponding index in the matrix (there is a flag to remember that).

Creating jobs and vehicles with associated locations is the responsibility of the user and `input::add_[job|vehicle]` just "adds" them to the current problem description. If no index is provided for a job or vehicle start/end, then adding them will set the correct index (consistent with the location position used for matrix computing).

Issue #79, opened 2018-02-12
Same position services


braktar:
Hi,

By sending a jobs at the same place than a depot I expect to find it next to the start point, which is not the case. This job is not returned at all.

Problem send : 
`
{
    "vehicles": [
    {
        "id": 0,
        "start_index": 0
    }],
    "jobs": [
    {
        "id": 0,
        "location_index": 0
    },
    {
        "id": 1,
        "location_index": 1
    }],
    "matrix": [
        [0, 1],
        [1, 0]
    ]
}
`

And the associated result
`
{
    "code": 0,
    "summary":
    {
        "cost": 1,
        "computing_times":
        {
            "loading": 0,
            "solving": 0
        }
    },
    "routes": [
    {
        "vehicle": 0,
        "cost": 1,
        "steps": [
        {
            "type": "start"
        },
        {
            "type": "job",
            "job": 1
        }]
    }]
}
`

I agree there is nothing to perform, but it the case of the CVRP, it could become an issue.
Every send jpb should be in the solution even if it's useless from the resolution point of view

jcoupey:
This is not a solving issue but rather related to parsing, as it only occurs when using `location_index` with a custom matrix. This input in the same situation without matrix

```
{"jobs":[{"location":[5,47],"id":0},{"location":[6,48],"id":1}],"vehicles":[{"start":[5,47],"id":0}]}
```

 does work as expected.

braktar:
Thanks,

I've done quite the same trick by my side, extending the matrix by duplicating the row and column from the jobs with the same location_index.

I could probably filter the services by my side as long as the clustering heuristic CVRP is not released :)

jcoupey:
The current status is that `location_index` should be unique to keep out of trouble. As you already noticed, adding rows and columns to extend the matrix and avoid duplicate `location_index` works fine, even if this is just a workaround.

What happens is that indices are stored in a `std::set` in `input`, so jobs with a duplicate `location_index` are simply not passed on to `tsp` for solving. I didn't see that coming when merging #59. There is more to it than just making sure we keep duplicates, as part of the code relies on a mapping between the index and the actual job which does not make sense anymore if we allow duplicate index.

jcoupey:
As far as I can tell, the changes currently in branch `fix/duplicate-index` should fix the problem for instances with duplicate `location_index` values. Hopefully nothing is changed for other instances, be it with unique `location_index` or using OSRM.

[...]

PR #81, opened 2018-02-13
Handle duplicate location_index

jcoupey:
## Issue

#79 

## Tasks

 - [x] get rid of mappings in `input::_index_to_job_rank` and `input::_index_to_loc_rank`
 - [x] store indices in a `std::vector` to keep jobs with duplicate `location_index` and pass them to `tsp` for solving
 - [ ] review

### DIFF
--- a/src/problems/tsp/tsp.cpp
+++ b/src/problems/tsp/tsp.cpp
@@ -307,9 +307,7 @@ solution tsp::solve(unsigned nb_threads) const {
   if (_has_start) {
     // Add start step.
     assert(current_sol.front() == _start);
-    auto rank = _input.get_location_rank_from_index(
-      _tsp_index_to_global[current_sol.front()]);
-    steps.emplace_back(TYPE::START, _input._locations[rank]);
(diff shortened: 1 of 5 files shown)


jcoupey:
There is still a problem when we rely on `start_index` to decide on `tsp::_start` value here:

https://github.com/VROOM-Project/vroom/blob/50042e719879cd6699630f853f92a3401e181d2e/src/problems/tsp/tsp.cpp#L34-L39

because we just pick the first match which could very well correspond to a job with same index if one decides to define vehicles *after* jobs (not the case in current json parsing). I ported the dummy example in #79 adjusting the code in `libvroom.cpp`. Using `add_vehicle` before `add_job` is fine, but the other way round results in a wrong job id at:

https://github.com/VROOM-Project/vroom/blob/50042e719879cd6699630f853f92a3401e181d2e/src/problems/tsp/tsp.cpp#L337

Job and vehicle start at index 0 are mixed up and in the output the job gets whatever id the vehicle had.

We need to retain the type of location along with ids in `input` to avoid that.

Issue #83, opened 2018-02-14
Expose a function to set the matrix from C++


jcoupey:
We hit a segfault in `input::check_cost_bound()` (computing of an upper bound for solution cost to prevent overflows) when using `add_job` and `add_vehicle` from C++ and manually setting the input matrix.

Steps to reproduce:
1. build `libvroom` minimal example (in `examples` folder)
2. run `./libvroom-example` and retrieve the matrix computed by OSRM
3. manually define the very same matrix (by-passing the call to OSRM):

```diff
--- a/examples/libvroom.cpp
+++ b/examples/libvroom.cpp
@@ -23,6 +23,29 @@ int main() {
   input problem_instance(std::move(routing_wrapper),
                          true); // Query for route geometry after solving.
 
+  matrix<cost_t> matrix_input(4);
+  // manually define matrix_input[i][j] values
+  problem_instance._matrix = matrix_input;
+
```

jcoupey:
Solved by #84

Issue #42, opened 2016-09-10
Build vroom as library


frodrigo:
It will be interesting to build vroom as library containing only the solver and have the main, tsplib and osrm loaders in a command linked with the lib.

jcoupey:
I can see how useful this could be for integration in other pieces of software. At the moment, a loader could build a problem from a string (or file name) and solving would return a string (or a rapidjson object).

I think a proper library would require to expose more than just this. For example a problem object that could be directly populated, and also an object to represent a solution and use it directly after solving.

frodrigo:
The idea is to use it as a native library in an Android application (@fijemax)

jcoupey:
This sounds cool. But how do you plan to re-use the solution output, just parse it again in the client application?

frodrigo:
The idea with a library if to share objects not exchanging and parsing files.

jcoupey:
> to share objects

Precisely what I mean: we can't do this right know. At the moment, there is no "solution" object that could be shared, only code [reaching out all over the place](https://github.com/VROOM-Project/vroom/blob/release-v1.0.0/src/utils/logger.cpp) to write the expected `json` to output.

 I think this should be considered in the light of #44 (I just filled this issue to write down things I had in mind before).

jcoupey:
This is probably getting nearer with the refactor for #44, as involved objects (locations, jobs, vehicles, routes, input problem, solution etc.) now have a proper data representation.

sashakh:
Hi,

I'm using vroom as library, adding such simple makefile patch:
```
diff --git a/src/makefile b/src/makefile
index c729fcf..0ff069f 100644
--- a/src/makefile
+++ b/src/makefile
@@ -10,6 +10,7 @@ LDLIBS = -lboost_system -lboost_regex -lboost_log -lboost_log_setup -lpthread -l
 
 # Using all cpp files in current directory.
 MAIN = ../bin/vroom
+LIB = ../lib/libvroom.a
 SRC = $(wildcard *.cpp)\
                        $(wildcard ./algorithms/*.cpp)\
                        $(wildcard ./routing/*.cpp)\
@@ -34,11 +35,14 @@ endif
 OBJ = $(SRC:.cpp=.o)
 
 # Main target.
-all : $(MAIN)
+all : $(MAIN) $(LIB)
 
 $(MAIN) : $(OBJ) main.o
        $(CXX) $(CXXFLAGS) -o $@ $^ $(LDLIBS)
 
[...]

PR #84, opened 2018-02-14
Fix check_cost_bound segfault

jcoupey:
## Issue

#83 

## Tasks

 - [x] expose an `input::set_matrix(matrix<cost_t>&&)` function
 - [x] adjust json input parsing to use it rather than manually setting `input::_matrix` (that should probably be private anyway)
 - [x] probably narrow `_max_cost_per_line` and `_max_cost_per_column` to the `check_cost_bound` scope
 - [x] update `libvroom.cpp`
 - [x] review


krypt-n:
You probably know this already, but it crashes because `_max_cost_per_line` is an empty vector. If the matrix is manually set in the current implementation, it has to be set manually as well.

jcoupey:
@krypt-n yep, I just forgot about that (and [my own comment](https://github.com/VROOM-Project/vroom/issues/42#issuecomment-341048281)) when filling the issue...

What we actually want is a single function to set the matrix that removes the need to take care of all the `_max_cost_per_*` stuff.

jcoupey:
Just updated tasks list and issue name accordingly...

Issue #90, opened 2018-03-09
Skills


jcoupey:
Now that we start supporting multiple vehicles with capacity constraints, we want to be able to restrict job service to some vehicles (think jobs requiring specific vehicle characteristics or driver skills).

Possible API sketch: jobs and vehicle get a new `skills` array which is just a set of integers (end users have to decide on their own internal representation). Job skills are mandatory, i.e. a job can only be served by a vehicle whose `skills` array contains **all** of the job skills. In other words job `j` is eligible to vehicle `v` iff `j.skills` is included in `v.skills`.

Ideally those should be optional (absence means no restriction), but like capacities we can enforce that they are either not provided, either provided for all jobs and vehicles.

jcoupey:
For API consistency, we probably want to have that in the TSP code too. No big deal, as it just means ruling out jobs that are not compatible with the one vehicle and have them show up as `unassigned`.

jcoupey:
Done in #96 (except for the TSP part, see #97).

PR #96, opened 2018-03-19
Handle skills

jcoupey:
## Issue

#90  

## Tasks

 - [x] Add skills as job/vehicle member and in ctor
 - [x] Parse skills in json input
 - [x] Filter compatible jobs as candidates in CVRP clustering
 - ~~[ ] Handle TSP case~~
 - [x] Update libvroom example
 - [x] Update docs
 - [ ] review


jcoupey:
Edit: ruled out the "handle TSP case" part as it should be dealt with as part of #97.

Issue #98, opened 2018-03-27
Instance rejected prematurely


fonsecadeline:
Hi, 

When trying to solve the following instance, I get the error message : "Too high cost values, stopping to avoid overflowing".

```
{"vrp":{"points":[
{"id":"depot","location":{"lat":43.698883,"lon":7.18244}},
{"id":11578,"location":{"lat":"43.742798","lon":"7.4241676"}},
{"id":11579,"location":{"lat":"43.740894","lon":"7.4177237"}},
{"id":11744,"location":{"lat":"27.241438","lon":"6.2586102"}}],

"vehicles":[
{"id":"A101","start_point_id":"depot"}],

"services":[
{"id":11578,"type":"service","sticky_vehicle_ids":["A101"],"activity":{"point_id":11578}},
{"id":11579,"type":"service","sticky_vehicle_ids":["A101"],"activity":{"point_id":11579}}],

"configuration":{"resolution":{"duration":30000,"iterations_without_improvment":100,"initial_time_out":20000,"time_out_multiplier":2}}}}
```

Indeed, point with ID = 11744 has an infinite distance defined in our preprocessing. However, there is no service at this point. Therefore, this infinite distance should not be taken into account.

jcoupey:
The check your instance is failing on was introduced to prevent overflowing when "infinite distances" are added in custom matrices (see #50 for more background).

I guess you have a pre-processing step on the above input before feeding it to VROOM, so it would be helpful if you could post this example formatted following [the VROOM API](https://github.com/VROOM-Project/vroom/blob/master/docs/API.md#input).

fonsecadeline:
Here is the instance provided to VROOM : 

> {"vehicles":[{"id":0,"start_index":0,"end_index":0}],"jobs":[{"id":0,"location_index":1},{"id":1,"location_index":2}],"matrix":[[0,2545,2129,2147483647],[2434,0,653,2147483647],[2094,563,0,2147483647],[2147483647,2147483647,2147483647,0]]}

jcoupey:
Thanks. What happens is indeed that the upper bound computation operates on the whole matrix.

[...]

PR #100, opened 2018-03-29
allow big values in matrix if not used

fonsecadeline:
## Issue

https://github.com/VROOM-Project/vroom/issues/98

## Tasks

 - [ ] review

### DIFF
--- a/src/structures/vroom/input/input.h
+++ b/src/structures/vroom/input/input.h
@@ -56,6 +56,7 @@ private:
   std::unique_ptr<vrp> get_problem() const;
   void check_cost_bound() const;
   void set_vehicle_to_job_compatibility();
+  std::unordered_set<int> _matrix_used_index;
 
 public:
   std::vector<job_t> _jobs;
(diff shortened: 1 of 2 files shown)


jcoupey:
@fonsecadeline thanks for giving it a go! Currently taking a look.

src/structures/vroom/input/input.h
```diff
@@ -56,6 +56,7 @@ class input {
   std::unique_ptr<vrp> get_problem() const;
   void check_cost_bound() const;
   void set_vehicle_to_job_compatibility();
+  std::vector<int> _matrix_used_index;
```
jcoupey: As we just need to remember the set of used indexes, I would go for a `std::unordered_set` here. This would save the need to go through the vector to check for duplicates before insertion: the `if (std::find[...]` checks could be replaced by a simple `insert`.

fonsecadeline:
Hi,

I corrected my commit :) feel free to tell me if there is still something I need to change.

Issue #108, opened 2018-05-03
#include organization


krypt-n:
As mentioned in #86, some include-statements could be moved around to improve compile times.

Things that might be worthwhile to look at:

* [x] Move unnecessary includes from .h files to .cpp files.
* [ ] Forward declare more structs/classes.
* [x] Change include paths to be relative to the project root instead of the current working directory.

jcoupey:
Done in #109.

sashakh:
Good improvement indeed. But two small issues:
- it breaks compilation when libosrm exists
- it breaks compilation of examples

So two tiny fixed in PR #116

PR #109, opened 2018-05-03
Refactor/includes

krypt-n:
## Issue

#108

## Tasks

 - [x] Check that the project still builds with all important gcc/clang versions
 - [x] Try the other steps mentioned in #108 
 - [ ] review

## Results of this PR

Measured with `time make -j1`
Baseline:
`make -j1  60.08s user 4.69s system 99% cpu 1:05.09 total`

Move includes in util/:
`make -j1  60.45s user 4.60s system 99% cpu 1:05.33 total`

Move includes in routing/:
`make -j1  48.56s user 4.15s system 99% cpu 52.934 total`

Move includes in structures/: 
`make -j1  47.44s user 4.09s system 99% cpu 51.747 total`

Split typedefs.h:
 `make -j1  34.08s user 2.76s system 99% cpu 37.034 total`

Move includes in problems/ and algorithms/: 
`make -j1  33.08s user 2.39s system 99% cpu 35.608 total`

Before: `make -j4  90.05s user 6.39s system 358% cpu 26.922 total`
After: `make -j4  57.71s user 3.79s system 362% cpu 16.982 total`

That's a noticeable improvement compared to the current master. Note that the big gains come from keeping the rapidjson and osrm headers from being included everywhere.

### DIFF
--- a/src/algorithms/kruskal.cpp
+++ b/src/algorithms/kruskal.cpp
@@ -7,6 +7,13 @@ All rights reserved (see LICENSE).
 
 */
 
+#include <algorithm>
+#include <numeric>
+#include <vector>
+
+#include "../structures/typedefs.h"
+#include "../structures/abstract/edge.h"
+
 #include "./kruskal.h"
 
 template <class T>
--- a/src/algorithms/kruskal.cpp
+++ b/src/algorithms/kruskal.cpp
@@ -11,10 +11,10 @@ All rights reserved (see LICENSE).
 #include <numeric>
 #include <vector>
 
-#include "../structures/typedefs.h"
-#include "../structures/abstract/edge.h"
+#include "structures/typedefs.h"
+#include "structures/abstract/edge.h"
 
-#include "./kruskal.h"
+#include "algorithms/kruskal.h"
 
 template <class T>
 undirected_graph<T> minimum_spanning_tree(const undirected_graph<T>& graph) {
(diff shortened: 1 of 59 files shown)


jcoupey:
Thanks for giving this a go! That's a nice springclean and a great build speedup.

> Check that the project still builds with all important gcc/clang versions

Everything fine with gcc 5.4.0 and clang 3.8.0 on my side, except a couple extra warnings with clang. It doesn't look like your work might introduce problems with specific versions? I mean that if we find some, they probably already exist in current master.

> Try the other steps mentioned in #108

Feel free to keep going! ;-) Happy to merge any work in that direction too.

krypt-n:
There are a few opportunities to foward declare structs in vroom, mostly `matrix<T>`. I don't think we can gain anything by doing that right now and I don't want to forward declare just for the sake of it.

Regarding the project-root-relative includes: I pushed two commits that rewrite all includes to be project-root relative. I don't have a strong opinion on this, feel free to not merge them if they e.g. break your workflow.

jcoupey:
On forward declare: OK to not overdo it if there is little gain.

On project-root-relative includes: no strong opinion either but I feel that it makes the source worktree easier to grasp, especially when replacing includes with many `../`.

I'll be adding a couple commits, mostly for include orderings, then merge. Thanks again for submitting!

PR #142, opened 2018-07-23
Refactor/tsp speedup

### LINKED ISSUE #141: TSP heuristic performance improvements
I had some spare time and decided to look at the performance of the tsp heuristic and found some small changes that improve performance somewhat. I'm gonna describe the changes and their impact in the PR.

krypt-n:
## Issue

Closes #141 

## Tasks

 - [x] Review algorithm changes
 - [x] review

## Description

This PR is a series of patches that improve the performance of the TSP heuristic via some microoptimisations. The patches can be applied and reviewed individually, they do not build upon each other. I think the impact on code quality and maintainability is very small and the performance gains are worth including them.

origin/master:
```
../bin/vroom -i ../bin/d1291.json -t 1  1.93s user 0.12s system 99% cpu 2.051 total
{"cost":53459,"unassigned":0,"computing_times":{"loading":142,"solving":1799}}
```

Improve undirected_graph 003cb5883733ecb51ddcf16a247ea48451c134b5:
```
../bin/vroom -i ../bin/d1291.json -t 1  1.57s user 0.10s system 98% cpu 1.691 total
{"cost":53364,"unassigned":0,"computing_times":{"loading":153,"solving":1430}}
```
The getter copied the `std::list`s anyway, so we avoid their creation cost in the constructor.

Inline edge getters f687ed4:
```
../bin/vroom -i ../bin/d1291.json -t 1  1.53s user 0.10s system 98% cpu 1.658 total
{"cost":53364,"unassigned":0,"computing_times":{"loading":149,"solving":1389}}
```

`get_weight()` showed up in some profiles since it is used to sort the edges by weight for Kruskal's algorithm.

Remove bounds checks in relocate 0ae260b:
```
../bin/vroom -i ../bin/d1291.json -t 1  1.46s user 0.09s system 99% cpu 1.560 total
{"cost":53364,"unassigned":0,"computing_times":{"loading":150,"solving":1329}}
```

Adding shortcut to relocate 696aa4b:
```
../bin/vroom -i ../bin/d1291.json -t 1  1.38s user 0.09s system 99% cpu 1.476 total
{"cost":53364,"unassigned":0,"computing_times":{"loading":139,"solving":1254}}
```

I think this change does not change how the relocate-operator works as long as the triangle inequality holds in the matrix. (However, since I am quite aware that it does not hold on e.g. data that OSRM returns, I have not touched the code below to assume that).

Improve datastructures in munkres c545796 :
```
../bin/vroom -i ../bin/d1291.json -t 1  1.37s user 0.08s system 99% cpu 1.454 total
{"cost":53364,"unassigned":0,"computing_times":{"loading":151,"solving":1223}}
```

The implementation of Munkres' Algorithm does work on all indices of a dense matrix. Thus we don't really need sets and maps and could use vector everywhere.

Replace ordered with unordered set 9154fda :
```
../bin/vroom -i ../bin/d1291.json -t 1  1.33s user 0.08s system 99% cpu 1.412 total
{"cost":53362,"unassigned":0,"computing_times":{"loading":159,"solving":1171}}
```

The tree operations were showing up in profiles.
Note: this commit changes the found solutions since munkres relies on the ordering sometimes

Replace labeling_y with vector b3f25d2: 
```
../bin/vroom -i ../bin/d1291.json -t 1  1.20s user 0.09s system 99% cpu 1.292 total
{"cost":53362,"unassigned":0,"computing_times":{"loading":153,"solving":1056}}
```

Replace labeling_x with vector 0674bde :
```
../bin/vroom -i ../bin/d1291.json -t 1  1.13s user 0.07s system 99% cpu 1.203 total
{"cost":53362,"unassigned":0,"computing_times":{"loading":153,"solving":969}}
```

Improve matrix loading time a bit d98a912:
```
../bin/vroom -i ../bin/d1291.json -t 1  1.05s user 0.09s system 99% cpu 1.143 total
{"cost":53362,"unassigned":0,"computing_times":{"loading":107,"solving":958}}
```
Not really related to the TSP heuristic, but eh well, it showed up in profiles.

## Result

I ran the TSP benchmark script with most of the problems from TSPLIB (I didn't yet run it on the larger ones).

`origin/master`:
```
,Gaps,Computing times
Min,0.09,7
First decile,1.2,11
Lower quartile,1.97,18
Median,2.96,70
Upper quartile,3.56,1419
Ninth decile,4.54,4905
Max,7.56,161364
```

This PR:
```
,Gaps,Computing times
Min,0.0,5
First decile,1.01,13
Lower quartile,1.97,19
Median,2.73,35
Upper quartile,3.59,837
Ninth decile,4.34,3362
Max,6.33,110023
```

### DIFF
--- a/src/algorithms/munkres.cpp
+++ b/src/algorithms/munkres.cpp
@@ -43,6 +43,7 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
     // Step 1.
 
     alternating_tree.clear();
+    std::vector<index_t> S_list;
     std::unordered_set<index_t> S;
     std::unordered_set<index_t> T_set;
 
@@ -52,6 +53,7 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
       ++unmatched_x;
     }
     S.insert(unmatched_x);
+    S_list.push_back(unmatched_x);
 
     // Saving relevant neighbors in equality graph in alternating_tree
     // and initializing slacks.
@@ -86,7 +88,7 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
         }
 
         // Update labelings
-        for (auto const& x : S) {
+        for (auto const& x : S_list) {
           labeling_x[x] = labeling_x[x] + alpha;
         }
         for (auto const& y : T_set) {
@@ -100,7 +102,7 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
             slack[y] = slack[y] - alpha;
 
             if (alternating_tree.find(y) == alternating_tree.end()) {
-              for (auto const& x : S) {
+              for (auto const& x : S_list) {
                 if (labeling_x[x] + labeling_y[y] == m[x][y]) {
                     alternating_tree.emplace(y, x);
                 }
@@ -129,7 +131,10 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
         // proceed to step 2.
         index_t matched_x = matching_y->second;
 
-        S.insert(matched_x);
+        auto p = S.insert(matched_x);
+        if (p.second) {
+          S_list.push_back(matched_x);
+        }
         T_set.insert(chosen_y);
 
         // Updating slacks.
--- a/src/algorithms/munkres.cpp
+++ b/src/algorithms/munkres.cpp
@@ -104,7 +104,8 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
             if (alternating_tree.find(y) == alternating_tree.end()) {
               for (auto const& x : S_list) {
                 if (labeling_x[x] + labeling_y[y] == m[x][y]) {
-                    alternating_tree.emplace(y, x);
+                  alternating_tree.emplace(y, x);
+                  break;
                 }
               }
             }
--- a/src/algorithms/munkres.cpp
+++ b/src/algorithms/munkres.cpp
@@ -20,7 +20,7 @@ std::unordered_map<index_t, index_t>
 minimum_weight_perfect_matching(const matrix<T>& m) {
 
   // Trivial initial labeling.
-  std::unordered_map<index_t, T> labeling_x;
+  std::vector<T> labeling_x(m.size(), 0);
   std::vector<T> labeling_y(m.size(), 0);
   for (index_t i = 0; i < m.size(); ++i) {
     T min_weight = std::numeric_limits<T>::max();
@@ -29,7 +29,7 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
         min_weight = m[i][j];
       }
     }
-    labeling_x.emplace(i, min_weight);
+    labeling_x[i] = min_weight;
   }
 
   // Initial empty matching.
@@ -58,10 +58,10 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
     std::vector<T> slack;
     slack.resize(m.size());
     for (index_t y = 0; y < m.size(); ++y) {
-      if (labeling_x.at(unmatched_x) + labeling_y[y] == m[unmatched_x][y]) {
+      if (labeling_x[unmatched_x] + labeling_y[y] == m[unmatched_x][y]) {
         alternating_tree.emplace(y, unmatched_x);
       }
-      slack[y] = m[unmatched_x][y] - labeling_x.at(unmatched_x) -
+      slack[y] = m[unmatched_x][y] - labeling_x[unmatched_x] -
                       labeling_y[y];
     }
 
@@ -87,7 +87,7 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
 
         // Update labelings
         for (auto const& x : S) {
-          labeling_x.at(x) = labeling_x.at(x) + alpha;
+          labeling_x[x] = labeling_x[x] + alpha;
         }
         for (auto const& y : T_set) {
           labeling_y[y] = labeling_y[y] - alpha;
@@ -101,7 +101,7 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
 
             if (alternating_tree.find(y) == alternating_tree.end()) {
               for (auto const& x : S) {
-                if (labeling_x.at(x) + labeling_y[y] == m[x][y]) {
+                if (labeling_x[x] + labeling_y[y] == m[x][y]) {
                     alternating_tree.emplace(y, x);
                 }
               }
@@ -136,7 +136,7 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
         for (index_t y = 0; y < m.size(); ++y) {
           T current_value = slack[y];
           T new_value =
-            m[matched_x][y] - labeling_x.at(matched_x) - labeling_y[y];
+            m[matched_x][y] - labeling_x[matched_x] - labeling_y[y];
           if (new_value < current_value) {
             slack[y] = new_value;
           }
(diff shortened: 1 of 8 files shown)


jcoupey:
Wow, thanks for taking the time to profile and work on changes, and also for making it easy to review step by step. Overall all nice improvements, just have a few comments/questions.

1. About 003cb58, I have a feeling that we could stick to using vector instead of list in the return of `get_adjacency_list` to avoid the vector to list copy. This would mean rewriting the work on the eulerian path in `christofides.cpp`. But then maybe replacing the constant-time `erase` used on a list by a vector equivalent would ruin the gain anyway, so it's probably best to leave it that way.

2. Question about commit f687ed4: I don't understand how moving the getters to `edge.h` changes anything from the compiler point of view. Is it related to the template?

3. About 0ae260b: I think this is a left-over from the time where `_edges` used to be an `unordered_map`, I was actually not aware of the overhead of `at` over `operator[]`. Not sure if you have a reason to not apply the same change to every occurrence of `_edges.at`? I think it would be safe to do so.

4. About 9154fda: we can also make the switch to `unordered_set` in `greedy_symmetric_approx_mwpm`, as long as order is consistent in the two `for` loops (don't know if this is guaranteed with `unordered_set`?). Returned matchings might change in case of cost ties, due to lookup re-ordering, but this would not change the overall greedy principle.

Also while we're cleaning up left-overs, we could remove redondant variables in this function:

```patch
--- a/src/algorithms/munkres.cpp
+++ b/src/algorithms/munkres.cpp
@@ -195,8 +195,6 @@ greedy_symmetric_approx_mwpm(const matrix<T>& m) {
 
   while (remaining_indices.size() > 0) {
     T min_weight = std::numeric_limits<T>::max();
-    index_t first_chosen_index;
-    index_t second_chosen_index;
     std::set<index_t>::iterator chosen_i;
     std::set<index_t>::iterator chosen_j;
     for (auto i = remaining_indices.begin(); i != remaining_indices.end();
@@ -207,14 +205,12 @@ greedy_symmetric_approx_mwpm(const matrix<T>& m) {
         T current_weight = m[*i][*j];
         if (current_weight < min_weight) {
           min_weight = current_weight;
-          first_chosen_index = *i;
-          second_chosen_index = *j;
           chosen_i = i;
           chosen_j = j;
         }
       }
     }
-    matching.emplace(first_chosen_index, second_chosen_index);
+    matching.emplace(*chosen_i, *chosen_j);
     remaining_indices.erase(chosen_j);
     remaining_indices.erase(chosen_i);
   }
```

jcoupey:
Results on all 78 TSPLIB instances on the machine I use for benchmarking:

`origin/master`

```
,Gaps,Computing times
Min,0.33,4
First decile,1.37,6
Lower quartile,2.03,8
Median,2.96,66
Upper quartile,3.72,1137
Ninth decile,4.41,19203
Max,7.56,3458318
```

This PR:

```
,Gaps,Computing times
Min,0.0,3
First decile,0.9,5
Lower quartile,1.78,6
Median,2.93,28
Upper quartile,3.54,466
Ninth decile,5.15,6785
Max,7.56,2245798
```

Total computing time is reduced by ~44%! :tada:

jcoupey:
Results on all 78 TSPLIB instances on the machine I use for benchmarking:

`origin/master`
```
,Gaps,Computing times
Min,0.33,4
First decile,1.37,6
Lower quartile,2.03,8
Median,2.96,66
Upper quartile,3.72,1137
Ninth decile,4.41,19203
Max,7.56,3458318
```

This PR:
```
,Gaps,Computing times
Min,0.0,3
First decile,0.9,5
Lower quartile,1.81,6
Median,2.94,28
Upper quartile,3.54,520
Ninth decile,5.13,13017
Max,7.56,2245798
```

Overall computing time is down by ~35%! :tada:

krypt-n:
> Question about commit f687ed4: I don't understand how moving the getters to edge.h changes anything from the compiler point of view. Is it related to the template?

It is not related to the template. On origin/master, when compiling the `std::sort` in kruskal.cpp, the compiler knows that there is a `edge::get_weight()`-method (since it is declared in the included edge.h header), but it does not know how it is implemented (since edge.cpp is compiled separately). Thus it has to use a function call to get the weight member of an edge class. With f687ed4 the compiler can see the actual implementation of the getter in the header and decide that it is small enough to be inlined into the comparison lambda (which can in turn be inlined into the instantiation of the std::sort template). This way the function call in a hot part of the code turns into a simple pointer dereference.

> Not sure if you have a reason to not apply the same change to every occurrence of _edges.at

No specific reason, just wanted to keep the changes small. Done in 90fcb06

> About 003cb58, I have a feeling that we could stick to using vector instead of list in the return of get_adjacency_list to avoid the vector to list copy

This is probably not a big performance issue right now, since the `eulerian_graph` is not a very dense graph.

> Also while we're cleaning up left-overs, we could remove redundant variables in this function:

Done in 2682f58

> don't know if this is guaranteed with unordered_set

It's not stated explicitly anywhere, but I think the documentation of `rehash` implies that the ordering only changes on `rehash`.

> we can also make the switch to unordered_set in greedy_symmetric_approx_mwpm

The matrix passed to `greedy_symmetric_approx_mwpm` does not seem to be very big in the christofides implementation. I can take a look at that if it starts to appear  in profiles for larger tsp instances.

jcoupey:
Another thing caught my eye: you report a worst gap of +6.33% with this PR, but I'm still getting the +7.56% (on instance `pr76`) on the machine I used to run the benchmarks. Investigating this a bit, I found out that this worst case is also reduced on my laptop. In fact most results actually differ between the other machine and my laptop (both running at d98a912). I checked the behaviour on `master` and some differences also appear, but to a much lesser extent.

My guess is that this is due to differences in the inner ordering behaviour from different standard library versions (the two machines run Ubuntu 16.04 and 18.04 respectively). And this has been all the more stressed in this PR by additional use of `unordered_set` over `set`. What do you think?

krypt-n:
I pushed two more commits to this branch:
067ac99 adds a `std::vector` for iteration over the elements of `S`. This means that the iteration order should no longer depend on the implementation of the hashmap but on the insertion order in the munkres algorithm.
8cb214f adds a break a inner loop since we only need to emplace something once. (This is closer to the inner loop on origin/master).

> What do you think?

I think we should try to produce consistent results across different machines and standard libraries. This allows reproducible benchmark results and the scenario where a job is assigned to a vehicle on one computer but unassigned on another computer should probably be avoided as soon as people start using vroom to plan stuff that involves real jobs. 

To continue this line of thinking, the result should probably be independent of the ordering of the input and maybe even independent of the ids.

src/algorithms/munkres.cpp
```diff
@@ -79,31 +79,32 @@ minimum_weight_perfect_matching(const matrix<T>& m) {
...
+            slack[y] = slack[y] - alpha;
 
-            for (auto const& x : S) {
-              if (labeling_x.at(x) + labeling_y.at(y) == m[x][y]) {
-                if (alternating_tree.find(y) == alternating_tree.end()) {
+            if (alternating_tree.find(y) == alternating_tree.end()) {
+              for (auto const& x : S_list) {
+                if (labeling_x[x] + labeling_y[y] == m[x][y]) {
                   alternating_tree.emplace(y, x);
+                  break;
```
jcoupey: Good catch for the `break` addition and loop re-ordering!

jcoupey:
On the matter of reproducible results across machines, this PR did not introduce the non-deterministic behaviours, only maybe made them a bit more noticeable. Handling this here seems out of scope so I'd be inclined to merge as is and ticket this question separately.

I this sounds good to you, could you please update the changelog by adding an item in an `[Unreleased]` entry? Don't worry about potential conflicts with current `master`, they'll be easy to fix upon merging.

krypt-n:
Done

jcoupey:
Shall I let you open a new ticket about reproducible solving based on your previous comment, or do you want me to open it?

Issue #134, opened 2018-07-05
CI builds


jcoupey:
Testing that changes don't break the build should really be automated. It's a PITA to check manually across several compilers, versions and settings (e.g. whether we're building against `libosrm` or not) and now and then some problems go unnoticed.

I did not really dig into the options so far, though it seems that Travis might be a good way to go.

jcoupey:
Builds should now be triggered for each commit on `master` and for the last one pushed on each branch that has a PR. Jobs settings include using gcc 6, 7 and 8, and clang 5. Use of clang 6 is currently turned off because the OSRM build is broken for `v5.18.0`.

Checking build status when `libosrm` is installed is possible with lots of boilerplate work like installing OSRM dependencies and building from source... In most situations this is a total waste of ressources and time. I used a hack to make it happen only whenever the `routing/libosrm_wrapper.*` files are modified in a PR. Those files are the only one that actualy include some `osrm/*` headers.

PR #154, opened 2018-09-04
Add CI builds with Travis

jcoupey:
## Issue

#134 

## Tasks

 - [x] Setup a basic Travis config
 - [x] Try to have support for clang 6
 - [x] Builds with `libosrm` installed
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ matrix:
           packages:
             - clang-4.0
             - libboost-all-dev
+            - libstdc++5
       env: COMPILER=clang++-4.0
 
     - os: linux
@@ -49,6 +50,7 @@ matrix:
           packages:
             - clang-5.0
             - libboost-all-dev
+            - libstdc++5
       env: COMPILER=clang++-5.0
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: cpp
 
+branches:
+  only:
+    - master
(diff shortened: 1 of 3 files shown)


jcoupey:
The Travis builds using gcc 6 and 7 are actually passing. The ones using clang 4 and 5 are failing in a strange way:

```shell
problems/cvrp/heuristics/clustering.cpp:188:23: error: no matching member function for call to 'erase'
        candidates[v].erase(init_job);
        ~~~~~~~~~~~~~~^~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/bits/stl_vector.h:1075:7: note: candidate function not viable: no known conversion from '__normal_iterator<const unsigned short *, [...]>' to '__normal_iterator<std::vector<unsigned short, std::allocator<unsigned short> >::pointer, [...]>' for 1st argument
      erase(iterator __position);
      ^
```

The same code compiles just fine with clang 5 on a laptop using Ubuntu xenial (16.04). As the Travis images are still under trusty (14.04), I suspect this is a problem with the `libstdc++` version.

jcoupey:
Thanks for the hint @krypt-n, definitely need to upgrade `libstdc++` for the clang builds.

jcoupey:
Following the proposed hack in [this SO post](https://stackoverflow.com/questions/46417416/travis-ci-seems-to-be-reading-from-the-wrong-stdlib) seems to change whatever `libstdc++` version clang is using... The build is no successful.

Issue #147, opened 2018-08-16
Early abort for relocate and Or-opt moves


jcoupey:
Commit 696aa4bc5273ddcb2279fb10fc46f2af80836eec introduced an early abort when evaluating a relocate step for the TSP local search, saving quite a few loops. It's based on the fact that we can be sure that a bunch of potential gains won't be better than the one already known, even without complete evaluation.

It seems like we could port this principle to the CVRP local search phase, both for `relocate` and `or-opt` moves, respectively [here](https://github.com/VROOM-Project/vroom/blob/master/src/problems/cvrp/local_search/local_search.cpp#L592) and [here](https://github.com/VROOM-Project/vroom/blob/master/src/problems/cvrp/local_search/local_search.cpp#L636).

The good part is that the necessary values are already pre-computed in `solution_state::node_gains` and `solution_state::edge_gains`.

jcoupey:
Done in #155.

PR #155, opened 2018-09-05
CVRP operators early abort

jcoupey:
## Issue

#147

## Tasks

 - [x] Add early stop for relocate and Or-opt operators
 - [x] Check timings + see if it's worth changing operators order
 - [x] Update `CHANGELOG.md`
 - [ ] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 - Update `clang-format` to 6.0 for automatic code formatting (#143)
 - Speed up TSP solving by over 35% on all TSPLIB instances (#142)
-- Speed up CVRP solving by over 65% on all CVRPLIB instances (#146)
+- Speed up CVRP solving by over 65%, then another ~8% on all CVRPLIB instances (#146, #147)
 - Keys `duration` and `arrival` are no longer optional in output and based on matrix values.
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


jcoupey:
6e5ada2 does most of the work here, and should not result in a any change in solution, except for very special situations where adding to a route actually make it cheaper (would require at least one triangular inequality violation for durations).

60ce360 is about doing all other operators checks before relocate and Or-opt, so that best known gains are potentially higher and we hit more early abort. The extra gain is quite small but there is no reason not to do it. In that case we might see solution changes (think gain ties between operators, and the first encountered is different). Practically solutions are all the same when running on all CVRPLIB benchmark classes (same cost **and** exact same solution). Rough timings are:

| `x` | `master` | 60ce360 | Gain (%) |
|:------------------|------------:|---------:|---------:|
| 0 | 0m48s | 0m43 | -10.4 |
| 1 | 1m44 | 1m29 | -14.4 |
| 2 | 3m4 | 2m52 | -6.5 |
| 3 | 4m23 | 4m1 | -8.4 |
| 4 | 8m33 | 7m51 | -8.2 |
| 5 | 16m3 | 14m54 | -7.2 |

Those are just obtained using `time run.sh [...]` and the comparisons for the fastest situations are probably not that meaningful. Yet this is roughly a 8% speedup, still nice to have for just a couple of checks and code blocks move.

Issue #123, opened 2018-05-25
Replace Munkres algorithm


jcoupey:
This is a dedicated ticket to experiment on https://github.com/VROOM-Project/vroom/issues/117#issuecomment-389442512.

TL;DR: it seems that what we really need is an implementation of [Edmonds' algorithm](https://en.wikipedia.org/wiki/Blossom_algorithm) in place of Munkres that currently produces sub-optimal results for part of the TSP heuristic.

Potential gains might include:

- speed-up the heuristic (the computational complexity looks lower)
- get better solutions from the heuristic
- maybe also speed-up the local-search convergence if starting from better solutions

krypt-n:
A thing to think about is that we are dealing with complete graphs that most likely satisfy the triangle inequality, so we could be able to cut some corners in the implementation.

jcoupey:
>  we could be able to cut some corners in the implementation

You mean for Edmond's? I did not actually got around taking a close look at the implementation, it seems far from intuitive at first sight. I actually recall considering it in the first place and falling back to Munkres that looked much simpler (although it's not really intuitive either).

nielsenko:
You could consider https://www.mimuw.edu.pl/~mucha/pub/mucha_sankowski_focs04.pdf.

nielsenko:
Regarding implementation of Edmonds's original maximum matching algorithm, this site does a fairly good job explaining it:
http://web.eecs.utk.edu/courses/spring2016/cosc494/notes/Edmonds/index.html

including actual code:
http://web.eecs.utk.edu/courses/spring2016/cosc494/notes/Edmonds/Edmonds.cpp

jcoupey:
Thanks @nielsenko for the references, I had quickly gone through some of them a while back but wasn't aware of all.

The thing is that this would only benefit to the TSP solving code. As it's now only a small subset of the problems we handle, the priority is rather low and I don't think I'll be able to take some time to look this up myself any time soon.

On the other hand, even it's not possible to include the available existing implementations in the code base, it might allow to do some quick tests to evaluate the potential gain. Let me know if you feel like giving this a try, I can provide a few pointers!

nielsenko:
I'll consider giving it a try, but I'm a bit rusty. Last time I slinged C++ code was before C++03 got any traction.

jcoupey:
> I'll consider giving it a try

Great!

[...]

PR #193, opened 2019-01-04
Try using blossom5 instead of munkres

nielsenko:
## Issue

This PR is an experiment to see if replacing munkres with blossom 5 will improve solution quality and performance of TSP. See #123 

This PR should not be merged to master as is.

## Tasks

 - [ ] review


nielsenko:
Unless I have made a blunder, I think we can exclude using _Blossom V_, as the TSP benchmarks suffers on all parameters.

jcoupey:
Closing based on results from https://github.com/VROOM-Project/vroom/issues/123#issuecomment-451433695

Issue #187, opened 2018-12-11
Query-time profile switch


jcoupey:
It would be nice to be able to decide on a routing setting with several profiles, and then offer a way to switch profile right in the json input, e.g. by providing a `profile` key to vehicles.

When using several `osrm-routed` profiles, this can already be done by adjusting the `-a` and `-p` flags, but this requires to now in advance what profile should be used. We could instead have a generic way to describe OSRM instances from the command-line perspective, something like:

```
vroom -a car:0.0.0.0 -p car:5000 -a bike:0.0.0.0 -p bike:5001 [...]
```

When using `libosrm`, we could rely on named shared memory datasets to be used as profile names, so this would be a follow-up from #181.

Notes:
- we should provide generic ways to setup the configuration directly from C++ for use with `libvroom`
- defining and using vehicles with several profiles in the same problem instance is out of scope for this ticket

jcoupey:
Couple thoughts on this.

1. The `-p` flag could probably be removed altogether as the port can be part of the address
2. The `-l` flag to toogle using `osrm-routed` or `libosrm` is not that clear, especially in the light of #192. We should remove it and have a `-r` flag to explicitly describe the routing engine to pick among:
  - osrm (default), uses `osrm-routed`
  - libosrm
  - ... (extensible at will)
3. The (undocumented) `-m` flag is replaced by a profile description in the API at the vehicle level, probably `vehicle.profile` defaulting to `car`. Providing an input file with vehicles that have different profiles is (at least for now) an error case.

jcoupey:
This landed with #196. The "Usage" wiki page should be updated after the next release to describe the new options usage.

Note: point 1. above is not possible as the port is directly used by `boost::asio` to handle the request. The updated syntax for `-a` and `-p` is the one described initially above.

Issue #181, opened 2018-11-23
Add support for multiple shared memory OSRM datasets


KieCap:
Since OSRM 5.17, multiple datasets can be loaded into memory at the same time, using:

```
$ ./osrm-datastore --dataset-name=mydata /path/to/data.osrm
$ ./osrm-routed --dataset-name=mydata --shared-memory
```

It'd be nice if VROOM had support for this, as extracting the entire planet for routing is a painstaking task!

It could also probably be used to load multiple profiles in memory and decide on the profile at the VROOM query level.

jcoupey:
It would be good to have this, first to be on par with what options OSRM has to offer. Then it would open up to more flexibility in the integration.

When using `osrm-routed` with several datasets - with or without shared memory - we already have a way to select what instance is triggered by either using different ports along with `-p`, or using the `-m` option for profile (requires to proxy the OSRM requests based on the infamous unused `{profile}` parameter in the OSRM API).

But none of this is doable right now when using `libosrm`. We could probably just reuse the `-m` value (currently unused with `libosrm`) so that it describes the dataset name on a per request basis.

jcoupey:
This landed with #196 and is related to #187, so using different datasets with `libosrm` is just like using different profiles.

Load the data using custom names:
```
./osrm-datastore --dataset-name=car /path/to/car_data.osrm
./osrm-datastore --dataset-name=foot /path/to/foot_data.osrm
```

Run VROOM with `libosrm`:
```
vroom -r libosrm [...]
```

Default behaviour with no change to the input file amounts to firing OSRM requests against the `car` profile/dataset. Hitting the `foot` profile requires to add a `profile: "foot"` key for **all** vehicles in input. Valid `profile` keys are determined by the dataset names loaded in memory with `osrm-datastore`.

@KieCap you can use this approach with any combination of profile/region that result in different datasets, so hopefully this addresses your initial concern of splitting the entire planet extraction.

Issue #195, opened 2019-01-09
Catch osrm::exception


jcoupey:
Trying to use `libosrm` from within VROOM when the routing data is not loaded into memory with `osrm-datastore` result in an error. The current catch is triggered by a `std::exception` [here](https://github.com/VROOM-Project/vroom/blob/186bd4d0f514457acdcfd138a9666a3e0da9b876/src/main.cpp#L143) because at the time of implementation `osrm::exception` was not exposed. This has been solved since OSRM v5.4 so it's now safe to handle this error using the more specific `osrm::exception`.

PR #196, opened 2019-01-09
Profile selection

jcoupey:
## Issue

This PR aims at adding a way to decide routing profile at query time (see #187). When using `libosrm`, this will be implemented via #181.

## Tasks

 - [x] Remove `-m` flag
 - [x] Add `-r` flag to decide on routing engine and remove `-l`
 - [x] Adjust `-a` and `-p` input parsing to make them profile-dependant
 - [x] Parse `profile` key for vehicles in input and check for consistency
 - [x] Set routing profile based on input
 - [x] Handle `libosrm` use-case with named shared memory datasets
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,20 @@
 
 ### Added
 
+- Optional `profile` key for vehicles to allow picking routing profile at query-time (#196)
+- `-r` command-line flag for explicit routing engine choice (#196)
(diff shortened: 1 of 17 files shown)


jcoupey:
Updating `libosrm` usage as described above means dealing with `osrm::exception`s anyway so #195 will also be fixed in the process.

Issue #202, opened 2019-01-18
Filter out "unrelated" vehicles during local search


jcoupey:
As a result of the skills distribution (or possibly via #201), it can happen that two vehicles in a problem are "unrelated", meaning they have a disjoint set of job candidates. In that case it does not make sense to try any of the local search operators involving those vehicles routes, because they're bound to be invalid anyway. Both vehicles routes should still be tested against routes from other "related" vehicles.

What currently happens in that situation is that the skill-based incompatibilities result in an early abort for those operators (no timing/constraint/cost check), but we still go through a bunch of unnecessary `for` loops.

Once the job/vehicles compatibilities are set, we should define a vehicle/vehicle compatibility matrix, and simply not set up a vehicle pair candidate for local search if both vehicles are "unrelated". Then:
- ensure the overhead is negligible using unaffected instances (all vehicles are "related")
- evaluate the gain on affected instances

jcoupey:
Done in #205

Issue #201, opened 2019-01-18
Filter out incompatible jobs and vehicles


jcoupey:
Job and vehicle incompatibilities can be defined in input using `skills`. Based on that, the local search phase is able to cut short a lot of neighbours exploration: no need for fancy time validity and cost checks if a job simply can't be moved in some targeted route. As a result, being able to filter out jobs and vehicles that are *a priori* incompatible can save us a lot of computing time.

Depending on the problem definition, skills may be the only restrictions, but in some cases incompatibilities could be derived from other constraints: think vehicle/job disjointed time windows, or very tight vehicle time windows that *de facto* restrict job candidates to the cheapest ones around.

Once `Input::_vehicle_to_job_compatibility` is set based on skills, we should go through it again. Whenever the insertion of a compatible job in an empty route for a vehicle is invalid, we should mark both as incompatible. Then:

- ensure the overhead is negligible using unaffected instances
- evaluate the gain on affected instances

jcoupey:
Done in #205

PR #205, opened 2019-01-21
Use incompatibilities

jcoupey:
## Issue

Covers #201 and #202 to speed-up local search by removing irrelevant neighbourhoods and tests.

## Tasks

 - [x] Add additional job/vehicle incompatibilities derived from constraints
 - [x] Don't include vehicle pairs for local search neighbourhoods if they don't share some job candidates
 - [x] Ensure the overhead is negligible using unaffected instances
 - [x] Evaluate the gain on affected instances
 - [x] Update `CHANGELOG.md`
 - [ ] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 - `-r` command-line flag for explicit routing engine choice (#196)
 - Support for multiple named datasets when using `libosrm` (#181)
 - Generic `vroom` namespace and several other specializations (#135)
+- Spot more job/vehicle incompatibilities derived from constraints (#201)
+- Filter out irrelevant local search neighbourhoods for vehicles with disjoint job candidates (#202)
 
 ### Changed
 
--- a/src/algorithms/local_search/local_search.cpp
+++ b/src/algorithms/local_search/local_search.cpp
@@ -265,11 +265,14 @@ void LocalSearch<Route,
     best_ops[v] = std::vector<std::unique_ptr<Operator>>(_nb_vehicles);
   }
 
-  // List of source/target pairs we need to test (all at first).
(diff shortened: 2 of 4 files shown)


jcoupey:
I've been testing this PR with (big) instances that are particularly affected because of very tight TW constraints. Computing times reductions are really significant, while the solutions are always the same (the whole solving process is unchanged, we just cut short some lookups that are bound to fail).

I'm reporting the results on different instances for 74f2ea8 (only job/vehicle incompatibilities added as described in #201) and f859cad (also include vehicle/vehicle incompatibilities as described in #202).

### Instance 1

Rate of added extra job/vehicle incompatibility: 96.1 %
Rate of added extra vehicle/vehicle incompatibility: 66.3 %

### Instance 2

Rate of added extra job/vehicle incompatibility: 84.9 %
Rate of added extra vehicle/vehicle incompatibility: 5.2 %

### Instance 3

Rate of added extra job/vehicle incompatibility: 75.1 %
Rate of added extra vehicle/vehicle incompatibility: 0.9 %

### Computing times (s)

|  | `master` | 74f2ea8 | f859cad |
|:------------------|------------:|------------:|------------:|
| Instance 1 | 25.0 | 8.4 | 6.1 |
| Instance 2 | 25.5 | 13.4 | 10.9 |
| Instance 3 | 20.1 | 13.1 | 10.7 |

Of course this is just a quick test based on sample test-cases, but the tendency is clear. As expected the reduction is all the more significant if the number of spotted incompatibilities is higher.

For unaffected instances, this only adds a negligible overhead (the filtering is `O(V*V*J)` where `V` and `J` are the number of vehicles and jobs respectively). The additional time is a matter of a few milliseconds, way lower than the overall computing time.

Issue #231, opened 2019-05-12
Implicit instantiation of undefined template error for macos g++ compiler


iedmrc:
With macos xcode compiler:
```
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
I get the following errors while building the source:

```
c++ -MMD -MP -I. -std=c++14 -Wextra -Wpedantic -Wall -O3 -c algorithms/kruskal.cpp -o algorithms/kruskal.o
In file included from algorithms/kruskal.cpp:14:
In file included from ./algorithms/kruskal.h:13:
In file included from ./structures/generic/undirected_graph.h:17:
In file included from ./structures/generic/edge.h:13:
./structures/typedefs.h:42:19: error: implicit instantiation of undefined template 'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
const std::string DEFAULT_PROFILE = "car";
                  ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/iosfwd:193:32: note: template is declared here
    class _LIBCPP_TEMPLATE_VIS basic_string;
                               ^
In file included from algorithms/kruskal.cpp:14:
In file included from ./algorithms/kruskal.h:13:
In file included from ./structures/generic/undirected_graph.h:17:
In file included from ./structures/generic/edge.h:13:
./structures/typedefs.h:49:15: error: implicit instantiation of undefined template 'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
  std::string host;
              ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/iosfwd:193:32: note: template is declared here
    class _LIBCPP_TEMPLATE_VIS basic_string;
                               ^
In file included from algorithms/kruskal.cpp:14:
In file included from ./algorithms/kruskal.h:13:
[...]

PR #232, opened 2019-05-12
implicit instantiation of undefined template error fixed for macos g++ compiler

iedmrc:
## Issue
#231 

Fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Implicit instantiation of undefined template error for macos g++ compiler (#231)
 - Parsing vehicle ids as `uint64_t` (#228)
 
 ## [v1.4.0] - 2019-03-26
--- a/src/structures/generic/edge.h
+++ b/src/structures/generic/edge.h
@@ -9,7 +9,7 @@ Copyright (c) 2015-2019, Julien Coupey.
 All rights reserved (see LICENSE).
 
 */
-
+#include <string>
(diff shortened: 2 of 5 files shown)


jcoupey:
Thanks for the PR!

Including `string` in `src/structures/typedefs.h` does make sense to fix the build errors you're experiencing. But is that not enough? I don't get the need for another inclusion in `src/structures/generic/edge.h`, `src/structures/generic/matrix.h` and ` src/structures/vroom/raw_route.h`?

iedmrc:
Ah yes. I got errors with this order:
src/structures/generic/edge.h
src/structures/generic/matrix.h
src/structures/vroom/raw_route.h
src/structures/typedefs.h

So I added `#include <string>` one by one to all in order to solve the problem 😄 
I can push again with only `src/structures/typedefs.h` included `#include <string>` if you say?

jcoupey:
> I can push again with only src/structures/typedefs.h included #include <string>

If it's enough to solve your compilation problem, then yes please! Also it would be great if you could add a line to the Unreleased/fixed entry in the changelog with the issue number.

vishalmhjn:
> Thanks for the PR!
> 
> Including `string` in `src/structures/typedefs.h` does make sense to fix the build errors you're experiencing. But is that not enough? I don't get the need for another inclusion in `src/structures/generic/edge.h`, `src/structures/generic/matrix.h` and ` src/structures/vroom/raw_route.h`?

This worked for me

PR #236, opened 2019-05-23
http wrapper class

### LINKED ISSUE #224: Add http wrapper class
We currently have two routing classes that use http requests to get routing results: `OrsHttpWrapper` and `RoutedWrapper` respectively for openrouteservice and osrm-routed.

Exact request syntax and http method differ, but both are sharing some duplicated code to handle the requests. We should move those parts to a dedicated `HttpWrapper` abstract parent class.

jcoupey:
## Issue

Fixes #224 

## Tasks

 - [x] Move http stuff using `boost::asio` to a dedicated class
 - [x] Test with OSRM (`osrm-routed`)
 - [x] Test with ORS
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- `HttpWrapper` class to factor code previously duplicated across routing wrappers (#224)
+
 ### Fixed
 
 - Implicit instantiation of undefined template error for macos g++ compiler (#231)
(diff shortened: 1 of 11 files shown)


jcoupey:
All http-related code for routing requests is now part of the `HttpWrapper` class. Routing wrappers that use http requests are renamed to `OsrmRoutedWrapper` and `OrsWrapper`. They inherit from both `RoutingWrapper` and `HttpWrapper` and override the `build_query` function member that is pure virtual within `HttpWrapper`.

It's really just moving code around so no change is expected.

I only tested it using `osrm-routed` so far, not ORS.

PR #247, opened 2019-06-28
Implement job priority

### LINKED ISSUE #246: Job priority
The current optimization objective has two levels:

1. Minimize the number of unassigned jobs
2. Minimize total travel time

So whenever the number of unassigned jobs is the same for two different solutions, we'll consider the one that has the lowest total travel time.

As a result, we'll tend to return as unassigned the jobs that are  "more expensive" (e.g. furthest or with higher amount). This is clearly not ideal in settings where jobs can have different levels of priority. Think "this job is due today", or "it would be nice if this jobs fits in but no big deal if not".

What we could do is add a new `priority` key for jobs in input, and then change the first objective above to minimize the sum of priority scores over unassigned jobs. The current behaviour would match the situation where all `priority` values are equal, which we could also use as default.

jcoupey:
## Issue

Fixes #246 

## Tasks

 - [x] Handle new `priority` key for jobs
~~- [ ] Adjust solomon-based heuristics~~
 ~~- [ ] Adjust clustering-based heuristics (CVRP only)~~
 - [x] Adjust cost in `ls::try_job_additions` so reinserting jobs during local search take priority into account
 - [x] See if more changes are required during the local search phase (dedicated operator?)
 - [x] Adjust `libvroom.cpp` example
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
After experimenting with adjusting the heuristics, I came to the conclusion that it's not the best approach. Basing job insertion (or even only route seeding) on priority does a great job making sure that the jobs with higher priority end up included in the end. But it has an important impact on heuristics efficiency wrt travel times. It even leads to regressions in situations where all jobs can be assigned: the impact is such that we end up with solutions that are much more expensive in term of travel times, and because the heuristics solutions are poorer, the local search phase is significantly longer.

The next step is to adjust job additions happening in `ls::try_job_additions`. It might be enough while not compromising travel times as it happens only upon local search changes and after reaching a local minima and removing some jobs.

Issue #254, opened 2019-07-09
Performance improvement opportunities


krypt-n:
Profiling reveals some easy to apply micro optimizations that improve the performance of vroom somewhat

jcoupey:
Landed with #255.

PR #255, opened 2019-07-09
Refactor/micro opts

krypt-n:
## Issue

#254 

## Tasks

 - [x] check the performance improvement on different benchmarks and different hardware
 - [x] review

I had some old commits lying around that apply some micro optimizations with the goal of improving the performance of vroom somewhat. I think slight loss in maintainability and modularity is justified by the gained speedup.

Benchmark results on the Solomon VRPTW instances:
```
master:
,Gaps,Computing times
Min,0.0,106
First decile,0.0,171
Lower quartile,0.84,264
Median,3.51,358
Upper quartile,5.49,914
Ninth decile,7.65,1218
Max,11.68,2230

this branch:
,Gaps,Computing times
Min,0.0,71
First decile,0.0,141
Lower quartile,0.84,193
Median,3.51,280
Upper quartile,5.49,632
Ninth decile,7.65,786
Max,11.68,870
```
The computed solutions are exactly the same. I do expect similar computing time gains on other benchmarks, but I have not measured it yet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - `HttpWrapper` class to factor code previously duplicated across routing wrappers (#224)
 
+### Changed
+
+- Speed up solving by 25% for CVRP and up to 30% for VRPTW benchmark instances (#255)
+
 ### Fixed
 
 - Implicit instantiation of undefined template error for macos g++ compiler (#231)
--- a/src/algorithms/local_search/local_search.cpp
+++ b/src/algorithms/local_search/local_search.cpp
@@ -121,6 +121,8 @@ void LocalSearch<Route,
                                                   routes,
                                                 double regret_coeff) {
   bool job_added;
+  std::vector<Gain> best_costs;
+  std::vector<Index> best_ranks;
 
   do {
     double best_cost = std::numeric_limits<double>::max();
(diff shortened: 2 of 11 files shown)


jcoupey:
@krypt-n thanks for the PR, this looks great, especially the worst-case computing time improvement. I'll run some experiments on my side with various VRPTW/CVRP benchmarks and report. Only TSP won't be affected because it has it's own logic.

I'm fine with all the changes, but I'd like to understand the `std::vector<bool>` part thoroughly. I've read your comment in 485d60934a6ec837c6e92db885eef0ed2de4018f but I don't quite understand what makes it faster in this case, despite the various casts when filling `_vehicle_to_job_compatibility` and accessing values from `vehicle_ok_with_jobs`?

Also is there a reason for not applying the same logic to the `_vehicle_to_vehicle_compatibility` vector?

krypt-n:
https://godbolt.org/z/LiwMb1 shows the difference between the two variants. `std::vector<bool>` is specialized to store 8 booleans in a byte and thus requires some bit-arithmetic to extract values. `std::vector<unsigned char>` is not and thus can extract values with a simple address dereference.

> despite the various casts when filling _vehicle_to_job_compatibility and accessing values from vehicle_ok_with_jobs

vroom far more often reads the values then it writes them. And as can be seen in the link above, the cast just results in a comparison with 0.

> Also is there a reason for not applying the same logic to the _vehicle_to_vehicle_compatibility vector?

`_vehicle_to_vehicle_compatiblity` didn't show up as important during profiling

jcoupey:
Thanks for the explanation. I'll report when I'm able to run some tests on my side.

jcoupey:
I did compare current master at 4486868 with this PR using 8c49bdb, running on my usual test machine with `-t 8` across all values of `-x` on all CVRP + VRPTW benchmarks.

- All solutions are strictly identical
- Average computing times for CVRP instances are consistently down by ~25%
- Average computing times for VRPTW instances are a bit further reduced (around 26-27% for Homberger, up to 30-33% for Solomon)

This is great!

@krypt-n I think we should mention this refactor in the changelog.

krypt-n:
That's great

I added a changelog entry with the last commit. Apologies for the delay

Issue #241, opened 2019-06-14
Support for mixing independent pickups and deliveries


vaibhavmehrotra94:
I am trying to use vroom for pickup and delivery scenario is it possible to use it for the same.

For Eg. :- Let's assume my vehicle has a capacity of 200 and I have jobs for delivery which have amount 100 each and pickup with amount 70. Now is it possible to configure VROOM's input in such a way that the truck is assigned to pickup 70 amount somewhere during the trip without breaching the vehicle capacity.

The provided JSON is what I tried to input for solving the above explained situation. Here I have taken job amount in negative for pickup's. (as for job's with id 2 and 4)

```{
  "jobs": [
    {
      "id": 1,
      "location": [77.6447257, 12.8911828],
      "time_windows": [[64800, 72000]],
      "service": 11880,
      "amount": [256]
    },
    {
      "id": 1,
      "location": [77.6447257, 12.8911828],
      "time_windows": [[64800, 72000]],
      "service": 0,
      "amount": [-82]
    },
    {
      "id": 2,
      "location": [77.6500367, 12.8892035],
      "time_windows": [[64800, 72000]],
      "service": 1620,
      "amount": [-205]
    },
    {
      "id": 3,
      "location": [77.644629, 12.8948467],
      "time_windows": [[28800, 43200]],
      "service": 3000,
      "amount": [226]
    },
    {
      "id": 3,
      "location": [77.644629, 12.8948467],
      "time_windows": [[28800, 43200]],
      "service": 0,
      "amount": [-262]
    },
    {
      "id": 4,
      "location": [77.6141978, 12.8590643],
      "time_windows": [[39600, 50400]],
      "service": 3420,
      "amount": [-247]
    },
    {
      "id": 5,
      "location": [77.6400817, 12.9048541],
      "time_windows": [[28800, 43200]],
      "service": 3540,
      "amount": [232]
    },
    {
      "id": 6,
      "location": [77.6178125, 12.853929],
      "time_windows": [[64800, 72000]],
      "service": 1680,
      "amount": [234]
    },
    {
      "id": 7,
      "location": [77.6375484, 12.9233757],
      "time_windows": [[28800, 43200]],
      "service": 1740,
      "amount": [127]
    },
    {
      "id": 8,
      "location": [77.64630199999999, 12.9004965],
      "time_windows": [[39600, 50400]],
      "service": 2400,
      "amount": [-156]
    },
    {
      "id": 9,
[...]

Issue #189, opened 2018-12-17
Same-route precedence constraints


nielsenko:
Would it be possible to assign a partial order to jobs, so that the sequences returned by vroom would respect this partial order?

Perhaps a rank could be assigned to jobs, and all jobs with lower rank must come before all jobs with higher rank. Jobs without rank should not be constrained by this.

This would disallow certain sequences, similar to how timeslots does today.

jcoupey:
Could you expand a bit on the use-cases you have in mind? What kind of situation would this allow to handle?

nielsenko:
Say you have a box of sorted letters. These are sorted in a static sequence, that you do not want to change, since it would require that you re-sort them. However, you also have a cage full of parcels that are unsorted, and that needs to be sorted anyhow. These you would like to mix with the letters on your route.

So the order of the letters is fixed (you can split the box between routes), but the order of the parcels is free. I envision using a "respect partial order" feature to address such a use-case.

Another use-case would be to ensure pickups happen before drop-offs. This would require something more flexible than the rank suggested above, ie. someway to specify that the pickup job _x_ goes before drop-off job _y_ (_x < y_). 

I was challenged with these use-case, by a representative from a postal service.

braktar:
In an industrial context, we usually met three types of ordering relations between jobs.

- Same route : Jobs have to be visited within the same route without regardless to the order.
- Order : Jobs have to be visited within the same route in a particular order, letting other jobs to be inserted between.
- Sequence : Jobs have to be visited within the same route in a particular order, and no job can be inserted between.

braktar:
Same Route : One of the first met of this constraint was for a customer delivering fresh or frozen food. If I remember well some parcels were packed together and only split at deliveries.

Sequence : This constraint is often met with press delivery when some city districts are outsourced and contractually fixed.

The Order constraint is a good start for pick-up and delivery, but my thought is that it could be limited when both missions can't be inserted within the same route. For example, if the route is already fulfilled with some other Jobs, the pickup can be inserted, but inserting the delivery require to unassign another job.
In the pick-up and delivery case, both jobs are directly linked, In the general case Jobs are not directly dependant with an Order relation (Unassign one, doesn't mean you can't perform the other jobs)

sinaabadi:
@jcoupey any news about adding this feature to vroom?

jcoupey:
> I have no idea right now on how and when this could be implemented

[...]

PR #262, opened 2019-08-08
Mixed pickups and deliveries

### LINKED ISSUE #197: Invalid solution with negative amount
When using a negative `amount`, an invalid solution can be reached and chosen for output, hitting an assertion when trying to format the solution:

```
vroom: ./utils/helpers.h:196: vroom::Solution vroom::utils::format_solution(const vroom::Input&, const RawSolution&): Assertion `amount <= v.capacity' failed.
```

This happens because during the local search there is no validity check wrt capacity for **removing** a job from a route. But removing a job can break the capacity restrictions if it holds a negative amount.

The rationale behind capacity values being [signed integers](https://github.com/VROOM-Project/vroom/blob/196597228c462c8f255c7584869e523fd536fad9/src/structures/typedefs.h#L31) is that I was thinking this could be useful to model some pick-up and drop-off situations. I'd like to keep this as is, so the fix would be to add validity checks for capacity to all local search operators upon job removals.

Note : using current `master` or `v1.3.0` with *positive values only* for capacities is safe from this bug.

### LINKED ISSUE #266: Cut down validity checks time
A bit of profiling on current master (e.g. on solomon instances) shows that most of the local search time ­- that is most of the overall time - is spent in various operators validity checks, while computing potential gain for operators does not take that long. This makes sense as validity checks are usually performed **prior** to potential gain checks, so they are the front-facing way of discarding the many potential move we explore.

Here is a [FlameGraph](https://github.com/brendangregg/FlameGraph) obtained with `master` at 42e901b on Solomon instance R203 : [42e901b_R203.svg](https://framadrop.org/r/lQ_ihnQqhn#U73XB6jqUgO0FcT5dyywLq9yG1xifFyddYwyXi6yYbs=). Interestingly we can see that `vroom::vrptw::CrossExchange::is_valid` and `vroom::vrptw::MixedExchange::is_valid` are taking a bigger part than other operators.

Now there are reasons why we should probably check for potential gain **before** validity most of the time:

1. In the light of #262 it appears that in some cases the complexity of the validity checks will increase, while computing potential gains will remain unchanged. So discarding operators based on gain should avoid costly checks altogether.
2. Because we lookup the best move between routes or inside a route, it means that most of the time we already have a good candidate handy (the best one so far). So as the search goes on discarding moves based on potential gains should occur increasingly often.

Now there is a reason why validity checks are currently performed before gain computing. In some cases validity for a move is dependant on orientation choices (think moving an edge to another route with the original orientation or reversed). In that case the actual gain depends on the result of the validity check. Nevertheless, we could start by computing and storing all potential gains based on orientation options, then use the max gain as a bound to decide whether it's worth checking for validity.

### LINKED ISSUE #267: Deprecate CVRP clustering heuristics?
The current clustering heuristics build job clusters for each vehicle, then run a single TSP per vehicle to get an initial set of routes. Because there is no notion of TW validity in the clustering process (that would be #139), it only applies to CVRP so far.

When used in a delivery-only context, this is all fine, but in the light of #262 this approach does not hold out-of-the-box anymore. Indeed we can come up with clusters that look globally fine as seen from the delivery/pickup sum point of view, but then we have no guarantee that the TSP solution will be OK wrt step-by-step capacity violations (think a situation where all pickups are done at the beginning of a route then all deliveries at the end).

So the question is to know whether we need to update the existing heuristics, or deprecate them altogether. I'd rather go with the second option because:

- having a CVRP-specific heuristic is a long-term burden for maintenance
- we already have a lot of parameter tuning options with the other heuristics, we can probably reach similar (or acceptable) quality/computing times on CVRP instances by just adjusting those while ruling out clustering approaches

This last point requires some investigation for a sound decision.

jcoupey:
## Issue

The scope of this PR is to implement support for mixing independent pickups and deliveries in a single problem (#241). It should also fix #197 along the way.

Here "independent" stands for "no precedence constraints involved" (as opposed to the discussion in #189). So this is "just" a matter of making sure all our capacity checks handle mixed scenarios.

## API

No big change required, amounts are already defined using `Capacity` which is an alias for `int64_t`. We should just make it clear that deliveries/pickups for jobs can be modeled using positive/negative capacities. Maybe just enforce that vehicle capacity values are unsigned integers. And perhaps report vehicle load at step level in output.

**EDIT** : https://github.com/VROOM-Project/vroom/pull/262#issuecomment-520786353

### Modelling

If capacities can be positive/negative in input, the easiest way is probably to split pickups and deliveries at job level. This could be achieved by replacing the current `Job::amount` by two members `Job::delivery` and `Job::pickup`.

Example: a job amount of `[4, 6, -2]` in input would result in a `[4, 6, 0]` job delivery and a `[0, 0, 2]` pickup at the `Job` object level.

**EDIT** : https://github.com/VROOM-Project/vroom/pull/262#issuecomment-520786353

### Maintaining relevant information

The crucial part here to avoid performance degradation is to keep constant-time lookups for all capacity checks. So going through a whole route part to decide if adding a job will break something before or after is out of question. We should be able to achieve this by storing relevant information upon route modifications (those do not happen very often compared to validity checks).

Say we have a route `s -> j_1 -> j_2 -> ... -> j_i -> ... -> j_n -> e` and pickup (resp. delivery) amount for job `j_i` is `p_i` (resp. `d_i`).

1. `Π_i = sum(p_k) for 1 <= k <= i` (all pickups up to `j_i` in the route)
2. `Δ_i = sum(d_k) for i < k <= n` (all pending deliveries after `j_i`)
3. Vehicle load at `j_i` is then `L_i = Π_i + Δ_i`
4. Max vehicle load up to `j_i` is `FL_i = max(L_k) for k <= i`
5. Max vehicle load after `j_i` is `BL_i = max(L_k) for i <= k <= n`

This is quite similar to what is currently stored in `SolutionState::fwd_amounts` and `SolutionState::bwd_amounts`, which are updated upon route modification through `SolutionState::update_amounts`. Warning: `LocalSearch::try_job_additions` currently has its own logic to update those values, it should be replaced to use the same function.

### Additions

Suppose we want to check whether it is OK to include:

- something with delivery `D` and pickup `P` (be it a job or bundle of jobs)
- between steps `i` and `i + 1`
- in a route for vehicle with capacity `C`

Then we have to check for `BL_{i + 1} + P <= C && FL_i + D <= C`.

### Removal

Should always be OK.

## Tasks

 - [x] Add `delivery` and `pickup` members for jobs
 - [x] Refactor `try_job_addition` to use `SolutionState::update_amounts`
 - [x] Store and maintain additional relevant information
 - [x] Move all amount-related members of `SolutionState` to `RawRoute`
 - [x] Add capacity check functions as `RawRoute` members.
 - [x] Adjust all capacity checks in Solomon-based heuristics
~~- [ ] Adjust all capacity checks in clustering heuristics~~
 - [x] Adjust `try_job_additions` capacity checks
 - [x] Adjust all local search operators capacity checks
 - [x] Add asserts on capacity/load upon final route construction
 - [x] Switch some gain and validity checks to fix #266 
~~- [ ] Fall back to simpler validity checks for delivery-only or pickup-only instances~~
~~- [ ] Update 2-opt early abort checks~~
 - [x] Add `pickup` and `delivery` keys at job level
 - [x] Deprecate `amount` keys in both input and output
 - [x] Deprecate CVRP heuristics to fix #267 
 - [x] Test on use-cases with pickups only and mixed scenarios
 - [x] Enforce that vehicle capacities are unsigned integers?
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
On second thoughts, we can't just keep the old `amount` values for jobs because it does not allow to model a situation where there are both pickups and deliveries at the same job. 

Example: say we have a delivery of `6` and a pickup of `5` (whatever metric) for a given job. We'd have to just pass the delta in the job `amount` as `1`, but we'd then be loosing mandatory information for capacity checks. The vehicle need to have enough room for a `6` delivery from the start up to that job, and enough room for a `5` pickup from that job on to the end of its route...

Solution: add two `pickup` and `delivery` keys at job level and deprecate `amount` keys in both input and output. To maintain previous behaviour before deprecation is effective, we'll probably have to treat an `amount` as a `delivery` with an empty `pickup`.

jcoupey:
So far I've been putting the new validity checks in place in the various local search operators without bothering too much about computing times. It should be mostly functional by now.

With the data stored, we're able to check efficiently if an addition to a route will incur a capacity violation before or after the insertion. What happens though is that going through the moved part is sometimes mandatory to make sure we don't have a capacity violation during the newly inserted part. As a result the computing time increase is "too high". I see a couple ways to deal with that:

1. Make sure we only pay the price of higher computing times when it's necessary (e.g. by storing a flag to know if we have a pickup-only or delivery-only instance, then falling back to the old cheaper checks in that case).
2. Adjust the overall approach tuning to reach a more acceptable trade-off between quality and computing times.
3. Reduce the number of validity checks altogether (ticketed in #266).
4. Solve some puzzling stuff I see while profiling (more on that next).

jcoupey:
More on point 4 from previous comment. I've been profiling executions with `perf` and doing some [flame graphs](https://github.com/brendangregg/FlameGraph) on Solomon instance R203 at 42e901b (parent commit in `master` for this PR) and 4d46ca4 (current PR HEAD):

- [42e901b_R203.svg](https://framadrop.org/r/lQ_ihnQqhn#U73XB6jqUgO0FcT5dyywLq9yG1xifFyddYwyXi6yYbs=)
- [4d46ca4_R203.svg](https://framadrop.org/r/yrumavp6ni#RSAoNWu90ezoEBz8KtKLRyKR5da2ZR+R+kLSUjw1bQM=)

I'm puzzled by:
- the huge depth of the second one compared to the first, this looks wrong
- the fact that we see **a lot** of `malloc` and `cfree@GLIBC_2.2.5` stuff appearing in the global view
- the changes observed when browsing to e.g. `vroom::vrptw::CrossExchange::is_valid` in both graphs. The one for 42e901b is totally expected (the usual TW checks with helper function, plus the calls to `vroom::cvrp::CrossExchange::is_valid` that consist mostly in `operator<=` checks for amounts). On the other hand for 4d46ca4, `vroom::cvrp::CrossExchange::is_valid` has a lot of stuff I don't get on top of the expected `vroom::RawRoute::*` checks, including some `cfree`, `malloc`, `operator new`, `delete`, etc.

One place where the latest start showing off is at f16aa71, the first check update:

```patch
$ git diff f16aa71~ f16aa71
diff --git a/src/problems/cvrp/operators/exchange.cpp b/src/problems/cvrp/operators/exchange.cpp
index 4f2a611..726a150 100644
--- a/src/problems/cvrp/operators/exchange.cpp
+++ b/src/problems/cvrp/operators/exchange.cpp
@@ -115,13 +115,19 @@ bool Exchange::is_valid() {
   bool valid = _input.vehicle_ok_with_job(t_vehicle, s_job_rank);
   valid &= _input.vehicle_ok_with_job(s_vehicle, t_job_rank);
 
-  valid &= (_sol_state.fwd_amounts[t_vehicle].back() -
-              _input.jobs[t_job_rank].amount + _input.jobs[s_job_rank].amount <=
-            _input.vehicles[t_vehicle].capacity);
-
-  valid &= (_sol_state.fwd_amounts[s_vehicle].back() -
-              _input.jobs[s_job_rank].amount + _input.jobs[t_job_rank].amount <=
-            _input.vehicles[s_vehicle].capacity);
+  valid &=
+    target.is_valid_addition_for_capacity(_input,
+                                          _input.jobs[s_route[s_rank]].pickup,
+                                          _input.jobs[s_route[s_rank]].delivery,
+                                          t_rank,
+                                          t_rank + 1);
+
+  valid &=
+    source.is_valid_addition_for_capacity(_input,
+                                          _input.jobs[t_route[t_rank]].pickup,
+                                          _input.jobs[t_route[t_rank]].delivery,
+                                          s_rank,
+                                          s_rank + 1);
 
   return valid;
 }
```

@krypt-n I suspect this might be related to the way I've been working with amounts for validity checks, any idea?

krypt-n:
The 5-argument `is_valid_addition_for_capacity` from the diff is now called `is_valid_addition_for_capacity_margins` if I am not mistaken.

```c++
bool RawRoute::is_valid_addition_for_capacity_margins(
  const Input& input,
  const Amount& pickup,
  const Amount& delivery,
  const Index first_rank,
  const Index last_rank) const {
  assert(1 <= last_rank);
  assert(last_rank <= route.size() + 1);

  auto first_deliveries =
    (first_rank == 0) ? current_loads[0] : bwd_deliveries[first_rank - 1];

  auto first_pickups = (first_rank == 0) ? Amount(input.amount_size())
                                         : fwd_pickups[first_rank - 1];

  auto replaced_deliveries = first_deliveries - bwd_deliveries[last_rank - 1];

  return (fwd_peaks[first_rank] + delivery <=
          capacity + replaced_deliveries) and
         (bwd_peaks[last_rank] + pickup <=
          capacity + fwd_pickups[last_rank - 1] - first_pickups);
}
```
I'd suggest the following changes which get rid of the allocations that are noticeable in the flamegraph:

* `auto first_deliveries` -> `auto& first_deliveries`. This avoids copying `Amount`s which is expensive since they are basically `std::vector`s. We can do this change since the lifetime of `current_loads` and `bwd_deliveries` is longer than the lifetime of the reference and other threads don't write to them (hopefully).

* Replace `Amount(input.amount_size())` with a global `const Amount zero` that is initialized once at the beginning to the correct size. Could be a new member of `Input`. This avoids allocating a new zero-Amount every time we use it.

* Now that both of the possible values are not temporaries, we can change `auto first_pickups` to `auto& first_pickups`. This avoids copying the involved `Amount` objects.

The declaration `auto replaced_deliveries` does not need to be turned into a reference since this just constructs an `AmountDiff` object, which does not involve allocations.

The other `is_valid_addition...` methods would need similar changes to get the performance back into the same ballpark as before.

Hope that helps

jcoupey:
@krypt-n  thanks a lot for your input, I've been able to reduce allocations following your suggestions. I still see some in the [flame graph for the last commit](https://framadrop.org/r/XGdqGJ2c2t#LcJOSFA5s2+wtIkxQO0Qz4TKsx3c/4oM1rsDlF/UNaU=) but this is probably related to [this `Amount` object](https://github.com/VROOM-Project/vroom/blob/04e1c1c7b39bf8ca5130bc70ab6a92e9e0b22372/src/structures/vroom/raw_route.cpp#L154-L166) creation we can't really get rid of.

jcoupey:
As several other changes are related to the current work in this PR, I'm extending it's scope to also cover #266 and #267. I've updated the task list accordingly.

jcoupey:
The last few commits to address #266 are really promising on the computing times side. Average computing time on `master` before this PR (42e901b) for Solomon benchmark instances was 604ms. It went up to 2052ms (more than a 3x increase!) at 474d6bc with the increase of complexity in the various `is_valid` implementations across local search operators.
Now reducing the number of validity checks altogether bring it down to 489ms for latest commit (cb5a4c8). It means that:

- we should be able to handle mixed pickups and deliveries as a new feature with similar computing times (maybe even a bit lower)
- we'll probably even get a nice speed-up for delivery-only instances when falling back to simpler validity checks for capacity

PR #275, opened 2019-10-07
Extended `cost` description

### LINKED ISSUE #272: Cost vs Duration vs Service
I'm unsure what the difference between `cost`, `duration`  and `service` is. My current understanding based on the docs and examples is the following:

- `service` is time it takes to load/unload when servicing a job;
- `duration` is the time it takes for travelling to the job location;
- `cost` is the total cost for completing a particular job. i.e. `service` + `duration`;

This applies to both the summary context and the route context in the output.

However, with v1.4.0, the `cost` is equal to `duration` for both the summary and the job contexts.

I've had a quick look at the open/closed issues but did not find any information on this. I'm happy to provide a PR for clarifying this in the docs.

unguul:
Extended `cost` description to clarify that it is from the optimizations point of view which currently just happens to be equal to duration.

## Issue

Fixes #272 

## Tasks

 - [ ] review

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -176,7 +176,7 @@ The `summary` object has the following properties:
 
 | Key         | Description |
 | ----------- | ----------- |
-| `cost` | total cost for all routes |
+| `cost` | total cost for all routes, as seen from the optimization point of view, currently equal to `duration` as optimization is done on travel time |
 | `unassigned` | number of jobs that could not be served |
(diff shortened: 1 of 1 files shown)


jcoupey:
@unguul thanks for submitting this clarification to the docs. I just moved the explanation to the preamble notes to avoid duplicating the info for `summary` and `route`.

Issue #186, opened 2018-12-11
Handle breaks


jcoupey:
Now that we have support for timing constraints, we should handle drivers breaks. To some extent, there are already some options to do that:

1. Hard-code working hours using one vehicle per time window (possibly with the same vehicle id). Downside: requires to impose the break location as end/start of the time windows, offers no flexibility with regard to timing (e.g. won't allow a small break shift past the time window end to handle one more job).
2. Define the break as a regular job with service time and time window. This is better in term of flexibility but still requires a fixed location, and offers no guarantee that the "break job" won't end up unassigned.

I want to define a break as some kind of "location-less" job. It should be movable forward or backward in time depending on surrounding jobs to solve the flexibility problem described above.

On the API side, we could add a `breaks` array for each vehicle, containing objects with a `service` and `time_window` key (not sure if there is a point in specifying several time windows for a given break?).

As far as I can tell for the solving side, this would require to:

- fill empty routes with breaks in the `tw_route` ctor to make sure they're mandatory
- apply the job time margin logic to breaks in `tw_route` to allow temporal moves while more jobs are added
- add some cost evaluation trick so that reaching from or to a break costs 0
- make sure breaks stay in place during the local search phase

There is probably some tricky stuff involved for operators gain evaluation and validity checks.

jcoupey:
@krypt-n thanks for your input! The implementation sketch above does not take into account this kind of regulation subtleties. The difficulty with these is that the break length(s) and position(s) would have to be decided *dynamically* at every step depending on the route characteristics. So for every tiny route modification, we'd be introducing a whole new range of combinatorial choices.

The approach I described does assume that breaks are "hard-coded" in input which is kind of limiting, though already allowing temporal flexibility. But I think it's a good first step that's already enough for many use-cases.

jcoupey:
I just merged #303 so this should now be fully functional on `master`. Any feedback appreciated!

PR #303, opened 2020-02-24
Feature/breaks

jcoupey:
## Issue

This PR targets #186 

## Tasks

 - [x] Add a struct to describe a break
 - [x] Parse breaks in input
 - [x] Maintain a data structure storing break ranks relative to `TWRoute::route`
 - [x] Init breaks in all new empty routes
 - [x] Take breaks into account for `add` timing validity checks
 - [x] Take breaks into account for `add` route updates
 - [x] Take breaks into account for `replace` timing validity checks
 - [x] Take breaks into account for `replace` route updates
 - [x] Take breaks into account for `remove` timing validity checks
 - [x] Take breaks into account for `remove` route updates
 - [x] Adjust output
 - [x] Update `libvroom` example
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
After spending a significant amount of time fiddling with the first idea of actually inserting breaks as "special" jobs in `TWRoute::route`, I think this is not the right thing to do. The main interest was that all timing-related checks would work out of the box, but then we'd break the current meaning of the cost of an "edge" between two jobs in a route. If a break is involved, this cost would need to account for previous/subsequent jobs/breaks (sometimes at several level of distance), introducing a whole new layer of complex case checks whenever a travel time cost has to be evaluated. Considering this happens everywhere throughout the code-base, I fear we'd end up with an unmaintainable code.

On the other hand, if we don't include breaks directly in `TWRoute::route` but somehow remember where they stand, then the whole cost-checking thing can remain unchanged. This would shift the touchy part to the time-related checks and would require to:

- maintain a data structure storing break ranks relative to `TWRoute::route`
- take breaks into account for all timing validity checks (`TWRoute::is_valid_addition_for_tw` and `is_valid_removal`)
- take breaks into account for all route updates (`TWRoute::add`, `remove` and `replace`) and update breaks data

Also in case of several breaks for a vehicle at a given rank (think an empty route at the beginning of the heuristic), we need to account for the insertion position of new jobs with regard to breaks. The best way to do this is probably to expose this position to the above validity checks and update functions.

Issue #307, opened 2020-03-03
Add Dockerfile


nilsnolde:
I'd like to PR a Dockerfile. It's fairly useless for `vroom` itself, however, if this acts as a base image for another `vroom-express` docker image, it would be very interesting. I think it's lowering generally the barrier to just give it a quick spin. For ORS we're also using Docker to deploy vroom in production actually. So both `vroom` & `vroom-express` run on the same server, which is fine for most use cases I think. Would be nice to have it upstream:) See PR in a few mins.

jcoupey:
Hi @nilsnolde, thanks for the input! I don't really use docker myself but I can see how many people would find it easier to go through docker to deploy the stack.

Before diving into the technical details you raise in #308, I'll start here with a few general questions:

1. How different would this be from, say https://github.com/iedmrc/vroom-docker?
2. Once the image creation process is setup, what would be the overhead in term of project maintenance? Asking mostly because I suspect we'd then get a lot of troubleshooting related to docker usage and setup itself (at least this is what I see happening on other projects bugtracker).
3. How does the routing layer come into the picture? People using a docker image would expect everything to work out of the box, but it won't unless they actually setup an OSRM or ORS stack aside.

nilsnolde:
Hi @jcoupey ,

1. There's a lot of advantages of having the Dockerfile(s) in the source repo(s) itself: 
- it's not a monolith like the Dockerfile you pointed to, i.e. `vroom` would be its own image, which `vroom-express` can build on and one can drag in any router one wants (and is supported) via `docker-compose`
- it's super flexible when it comes to releases, i.e. each release would automatically create its own Docker image and be tagged accordingly so people don't have to work off `master` as in the above Dockerfile
- it creates more trust IMO, if the Dockerfile is in the source repo, then one knows it's not gonna get abandoned unless the project is

Tbh, we could do it without committing to this repository and still retain most of the flexibility. Our @MichaelsJP [did that](https://github.com/gis-ops/docker-valhalla) for Valhalla and it became a major job to get it right while retaining all flexibility a "native" solution would have had.

[...]

PR #308, opened 2020-03-03
WIP: Add Dockerfile

nilsnolde:
Solves Issue # 307

As discussed in #307, a Dockerfile for Vroom itself is pretty useless. I'd see this as the base image for `vroom-express` to run everything with a single `docker run -dt -p 3000:3000 vroom-express`. In the end, I imagine the following:

- the `vroom` image is built multi-stage to keep the image size down; basically in a first stage it builds vroom and the second stage is based on a NodeJS image (for `vroom-express`) and simply copies from the first stage whatever it needs to run `vroom`
- the `vroom` image is hosted on Dockerhub with `latest` tag for `master` branch and other tags for releases
- `vroom-express` (I'd PR this soon) takes the second stage of the `vroom` image as base image

 Now, there's a few problems I'd like to address before this can reasonably happen:

- `vroom-express` would ideally create a release with every `vroom` release, always keeping the newest `vroom` release tag in the Dockerfile or else Docker tags need to be introduced which refer to `vroom` versions EDIT: actually that's not necessary, `vroom-express` wouldn't need to be on Dockerhub so one could specify whichever `vroom` version is preferred (and compatible)
- Currently it's not working yet for `vroom-express` as there's some libraries missing in the final image. Can you tell me which libraries `vroom` needs to just run (not build)? I'd like to keep an Alpine image, `boost-dev` is available. I tried the Ubuntu:18.04 image quickly but it was 10x the size of the `node:12.13.1-alpine` image.

I'd commit to keeping the Docker stuff up-to-date in both repositories, in case you're interested.

## Tasks

 - [ ] clear up final questions & get your approval;)
 - [ ] review

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+*.o
+*.d
+bin/
+lib/
+scripts/
+libvroom_examples/
+docs/
+.github/
+.git/
+Dockerfile
+README.md
+LICENSE
+CHANGELOG.md
+CONTRIBUTING.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM ubuntu:18.04 as build
+LABEL maintainer="nils@gis-ops.com"
+
+WORKDIR /vroom
+
+RUN apt-get -y update; apt-get install -y \
+    build-essential \
+		g++ \
+    libssl-dev \
+		libboost-all-dev \
+		pkg-config
+
+COPY . .
+
+RUN make -C /vroom/src
+
+FROM node:12.13.1-alpine
+
+COPY --from=build /vroom/bin/vroom /usr/local/bin/vroom


jcoupey:
> the `vroom` image is built multi-stage

I don't get the need for this. Is it not possible to have the NodeJS stuff come into play at the `vroom-express` level only?

> latest tag for master branch and other tags for releases

Is it really mandatory to build an image for each new push to `master`? For most use-cases it feels like having an image per release would be enough?

> vroom-express wouldn't need to be on Dockerhub

You mean users would clone the `vroom-express` repo and run a docker command from there?

The library used are listed [in the makefile](https://github.com/VROOM-Project/vroom/blob/02126a885994e859bfeaf959d263933000850160/src/makefile#L9).

I think we should benchmark the use of Alpine over e.g. Ubuntu. The latter uses `glibc` whereas Alpine relies on `musl` C library. First I recall this used to cause performance troubles for the OSRM docker images (see [some OSRM issues](https://github.com/Project-OSRM/osrm-backend/issues?q=is%3Aissue+docker+alpine+musl+library+performance+) for details). Second it's probably calling for trouble to base the image on very different environment (libc + dependencies). I've stumbled on [this related post](https://dev.to/asyazwan/moving-away-from-alpine-30n4).

nilsnolde:
Ah you know what.. Your questions made me doubt my thoughts here as well.. Let's just do a Dockerfile for `vroom-express` which builds `vroom` from source in its own Ubuntu-based image and a flexible build parameter for the desired `vroom` version. That build parameter can vary for different `vroom-express` releases, so it stays compatible. I don't know why I wanted to make it so complicated.. It really doesn't make sense to build `vroom` in Docker.. Sorry for the time waste..

Agreed?

jcoupey:
No time wasted if we come up with a simpler solution! ;-)

Also this sounds like a good starting point and we can still adjust things afterwards if needed.

Another question: rather than keeping things going in the `vroom-express` repo, wouldn't it be more convenient to have a dedicated repo (`vroom-docker` or something) for all docker-related stuff? The main advantage I see would be to have a dedicated place for all docker-related stuff: documentation, bugtracker etc.

nilsnolde:
Detaching the Dockerfile entirely from the repo is of course possible. Then you'll have to `git clone` it all in there and make sure to reference the right `vroom` version AND the right `vroom-express` version. We can do that too if you prefer that. Being dedicated to only Docker stuff is a valid point definitely. Yeah, maybe you're right, doesn't even make that much sense to put it only in `vroom-express`...

In that case, I'd create releases for the Docker repo to use hard-coded `vroom` and `vroom-express` releases. That sounds good as well. The release tags could be the same as the `vroom` release tags then with patch releases for `vroom-express` updates which are not relevant to `vroom` itself?

PR #369, opened 2020-07-28
Computing more flexible job earliest/latest dates

### LINKED ISSUE #330: Add flexibility for job time windows handling
Handling time windows for jobs works by storing additional data along routes, such as earliest and latest possible arrival dates at each step. Those are used to decide whenever a possible route change is feasible wrt timing constraints, and are updated each time a route is changed.

Currently, for a given job we [pick the first compatible time window](https://github.com/VROOM-Project/vroom/blob/c45754eef74b5e7537b718314a4c3d0795be0245/src/structures/vroom/tw_route.cpp#L268-L276) based on the earliest arrival candidate, then use it to update the new job earliest date **and it's latest date**. This is actually way too restrictive as we could avoid choosing the time window altogether and define a potentially looser latest date by picking another time window down the line.

As a result, in a situation with many multiple time windows, we're currently setting aside doable options that we consider as invalid due to the time window choice.

Consider for example jobs `A` and `B` whose respective `time_windows` arrays are: `[[8h, 12h], [14h, 16h]]` and `[[14h, 16h]]`. If the heuristic process starts by adding `A` to an empty route, then the matching earliest date will be `earliest_A = 8h` (or a bit more depending on travel time from the start) and because we picked the morning time window, it's latest date will be `latest_A = 12h` (or a bit less depending on travel time to the end). Then `A -> B` is fine, but `B -> A` is wrongly considered invalid because the new earliest arrival at `A` after `B` is necessarily after current `latest_A`. We should allow picking different time windows for earliest and latest date, in which case `latest_A` would be `16h` in the first place and `B -> A` would be valid.

As far as I can tell, there would be no drawback with this change (except a bit more fiddling to search the relevant time window for both values). No change in validity checks, no need to store job time window any more, it would just be decided upon route formatting when we truly set arrival times.

jcoupey:
## Issue

Fixes #330 

## Tasks

 - [x] Compute latest/earliest dates based on (potentially) different job TW
 - [x] Remove TW rank storage for jobs
 - [x] Lookup job TW in `format_route` to account for the situation of earliest and latest dates not belonging to the same TW
 - [x] Evaluate gains on ad-hoc benchmarks
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Support for string description of vehicles and tasks (#235)
 
+### Changed
+
+- Improved job earliest/latest dates handling internally (#330)
+
 ### Fixed
 
 - Documentation mismatch (#361)
--- a/src/structures/vroom/tw_route.cpp
+++ b/src/structures/vroom/tw_route.cpp
@@ -170,8 +170,13 @@ void TWRoute::fwd_update_earliest_from(const Input& input, Index rank) {
     // Back to the job after breaks.
     current_earliest += previous_service + remaining_travel_time;
 
-    const auto& next_TW = next_j.tws[tw_ranks[i]];
-    current_earliest = std::max(current_earliest, next_TW.start);
+    const auto j_tw =
+      std::find_if(next_j.tws.begin(), next_j.tws.end(), [&](const auto& tw) {
+        return current_earliest <= tw.end;
+      });
+    assert(j_tw != next_j.tws.end());
+
+    current_earliest = std::max(current_earliest, j_tw->start);
 
     assert(current_earliest <= latest[i]);
     if (current_earliest == earliest[i]) {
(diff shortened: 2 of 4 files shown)


jcoupey:
TL;DR the gain is really moderate and only for specific situations (multiple time windows in input), but still worth the trouble especially with so little changes to the codebase.

Longer version: I've checked results on specific situations where I was expecting this would make a difference. I also checked on a couple TW benchmarks with single time windows: no change in results, as expected.

## Scenario 1

### Setting

Typical weekly scenario where all jobs can be done on any day of the week (5 identical disjoints time windows). 20 randomly generated instances with each 4 vehicles / 200 jobs with a normal distribution in a regional-sized bounding box.

### Results

They vary across different exploration levels so I'm reporting average values.

- solution cost is strictly reduced for ~87% of the instances
- cost is reduced by 1.6% on average
- computing time is down by 18% on average

That last point what not directly expected but is a nice outcome. It's probably due to the fact that adding more flexibility in the earliest/latest dates allows to reach better solutions in the early stages so we then enjoy a faster overall convergence of the solving approach. This analysis is backed by the fact that low exploration levels see a bigger computing time reduction ratio.

## Scenario 2

### Setting

Typical morning/afternoon scenario where:

- 40% of jobs have `[8h; 12h]` and `[14h; 18h]` time windows
- 30% of jobs have a `[8h; 14h]` time window
- 30% of jobs have a `[12h; 18h]` time window

20 randomly generated instances with each 10 vehicles / 100 jobs (making sure everything gets assigned) with a normal distribution in a regional-sized bounding box.

### Results

The impact here is much less noticeable, as expected because there are at most 2 disjoint time windows at stake.

- solution cost is identical for ~19% of the instances and strictly reduced for ~54% of the instances
- cost is reduced by 0.4% on average
- computing time is down by 4% on average

## Scenario 3

### Setting

Same as 2 but using 300 jobs per instance so that some jobs are unassigned in most situations.

### Results

Due to unassigned jobs, the travel costs are not directly comparable any more, but the interesting point is that the number of unassigned jobs drops by around 7.7%. So the added flexibility also helps reaching solutions with more jobs included.

Issue #324, opened 2020-04-17
Reach better solutions using unassigned jobs


jcoupey:
The job [removal/re-insertion strategy](https://github.com/VROOM-Project/vroom/blob/625eebff8b41d9effcc46c6e77ad4ba3c39d332c/src/algorithms/local_search/local_search.cpp#L1361-L1372) applied after each local search step was primarily designed to improve the way already assigned jobs where handled across multiple routes. It can also help improve the solution with regard to unassigned jobs, but fails to do so in some situations where obvious better solutions are close (#319).

We should introduce a new local search operator that goes through each step for each route to evaluate the gain of removing it and inserting another unassigned job somewhere in the route. This would differ from other existing operators in that it might change the set of (un)assigned jobs and thus potentially change priority on top of travel time cost.

jcoupey:
Landed with #383.

Issue #319, opened 2020-04-11
Optimisation missing obvious solutions


romainrey:
Hi,

I am using vroom and my solutions are often sub optimal and I can spot right away a better solution if I look at a map.
I am using vroom with 1000-1500 jobs and 30 vehicles. But my time windows are such that usually only 5 jobs can be assigned to a vehicles. I also have priorities among this jobs, priorities going from 1 to 5.
The solutions usually becomes better once I limit the number of vehicles of solutions. But this is not doable for my real world solution.

I am reaching some limits of the optimisation or I am doing something wrong?
By suboptimal I mean
- Sometimes there is a doable solution that can have a higher sum of priorities
- Sometimes for the same number of points there is a shorter solution
________________________________________________

I am using vroom 1.5.0.
with this config file:
cliArgs:
  geometry: false # retrieve geometry (-g)
  threads: 3 # number of threads to use (-t)
  explore: 5 # exploration level to use (0..5) (-x)
  limit: '5mb' # max request size
  logdir: '/..' # the path for the logs relative to ./src
  maxlocations: 5000 # max number of jobs/shipments locations
  maxvehicles: 500 # max number of vehicles
  override: true # allow cli options override (-g, -t and -x)
  path: '' # VROOM path (if not in $PATH)
  port: 3000 # expressjs port
  router: 'osrm' # routing backend (osrm, libosrm or ors)
  timeout: 300000 # milli-seconds

__________________________________________________

I have tested it by putting an input with an other matrix, not computed from my local osrm but I still have problems.

__________________________________________________

I can provide a reproducible example, but I can't post it here because I don't want to expose my data... (just send me a mail at rom.rey06 at gmail.com and I can send it to you)

Thanks for your help!

jcoupey:
> Sometimes there is a doable solution that can have a higher sum of priorities
> Sometimes for the same number of points there is a shorter solution

First we provide no guarantee of optimality, so both of the above can definitely happen. What bothers me is if it does **and** we miss an obvious change that would improve the solution.

We have a range of local search operators that are applied in such a way that the provided solution should be free of some obvious *valid* changes: e.g. adding an unassigned job, moving a job from one route to another or within a route, exchanging two jobs from two different routes, etc.

So in order to fully see what you mean I'll have to look at examples to reproduce both situations with indications on what improvement we're missing.

romainrey:
Hi,

After discussing offline with Julien, here is a recap of the problem:

Use Case:
[...]

PR #383, opened 2020-08-28
Add new local search operator to exchange unassigned job

jcoupey:
## Issue

This PR aims at fixing #324

## Tasks

 - [x] Apply local search move based on priority increase first, then travel time gain
 - [x] Implement `UnassignedExchange` operator that tries to remove an existing job in order to onboard an unassigned job
 - [x] Use new operator in `LocalSearch::run_ls_step()`
 - [x] Early abort all priority-constant move lookups if we already have a move increasing priority for involved routes
 - [x] Discard `UnassignedExchange` moves involving jobs already assigned elsewhere
~~- [ ] Make `SolutionState::unassigned` an ordered `set` to browse unassigned candidates by decreasing priority~~
 - [x] Benchmark
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
Results on the problematic instance from #319 are encouraging (exploration level 4 and 5 are not reported as they reach the same solution as the one obtained with 3).

| Exploration level | 0 | 1 | 2 | 3 |
| :--- | ---: | ---: | ---: | ---: |
| **`master` priority score** | 44 | 148 | 152 | 152 | 
| **`master` duration** | 16137 | 25360 | 34915 | 34915 |
| **`e76e75b` priority score** | 152 | 152 | 159 | 159 | 
| **`e76e75b` duration** | 25232 | 25232 | 32330 | 28566 | 

So on this instance this PR reaches higher priority scores, on top of overall travel times that are usually lower.

From quick checks I've been running, it looks like on other benchmarks with unassigned jobs (and no notion of priority), this PR also helps including more jobs to some degree.

jcoupey:
I've reused a set of custom VRPTW benchmarks with various combinations of 200-job instances:

- single depot
- multiple depot
- start only, end only, different start and end (both identical or different across vehicles)

All of these with combinations of long/short planning horizon and loose/tight time-window constraint.

Because I want to test this PR in situations with many unassigned jobs, I reduced the number of vehicles to make sure it's way more difficult (often impossible) to handle all jobs.

| Exploration level | 0 | 1 | 2 | 3 | 4 | 5 |
| :--- | ---: | ---: | ---: | ---: | ---: | ---: |
| **`master` assigned job (%)** | 85.9 | 86.4 | 86.6 | 86.9 | 87.2 | 87.5 |
| **`master` average cost** | 23915 | 23897 | 23854 | 23869 | 23825 | 23810 |
| **`master` average computing time (ms)** | 150 | 269 | 465 | 587 | 1230 | 1965 |
| **`d28e144` assigned job (%)** | 87.4 | 87.7 | 88.0 | 88.2 | 88.4 | 88.7 |
| **`d28e144` average cost** | 23228 | 23213 | 23140 | 23129 | 23101 | 23071 |
| **`d28e144` average computing time (ms)** | 388 | 731 | 1266 | 1830 | 3148 | 4871|

Again the interesting thing is we now reach solutions with more jobs **and** at a lower cost on average. Of course the computing time does increase quite significantly but this is expected as we're adding loops across unassigned jobs in a situation with typically a lot of unassigned jobs.

On the other hand, I would expect the computing times to not change at all on instances where everything is easily included.

Issue #393, opened 2020-09-21
Favorize some vehicles over others for some jobs


romainrey:
Hi,

I am running into a case where some vehicles have some relations with some of the jobs.
Let's say that I have commercial sellers that are used to work with the clients of their zone.
I want to put an incentive for the commercial of the zone to go to his clients instead of the clients of the other commercial.
Skills does not work because it would be forcing him to go to his, while the idea here is more a penalization than a hard constrain. Indeed some time a commercial can go see the client of his colleague if his colleague has much more work than him.

What could work would be to allow priorities of jobs to be dependent of the vehicles. So for each job instead of having a unique priority it would have n priorities with n being the number of vehicles.
What do you think, is there any other workaround?

Ps:
I am already using priorities to incentive on urgency of the job. So my idea was to do priority -> priority + 2 if it is the commercial that knows the client.

Thanks!

jcoupey:
Adding this kind of soft constraint/penalization is something that usually is in conflict with the current solving approach and the way we set the optimization objective(s). See #263 for some background and discussions.

In this specific case, the whole notion of priority is currently handled as a "forefront" optimization objective, overall travel time being a secondary objective. In this prospect, I *think* the approach you describe would work for your special use-case.

On the other hand, I'm a bit circumspect at first on setting vehicle-dependant job priorities. First it would require to input much more data with a probably less intuitive API, then I'm not sure how we'd mix this with other ideas such as #359.

@romainrey if you want to go ahead and experiment with this idea, getting a minimal implementation to work would be quite straightforward because there are not that many places in the codebase where we deal with priorities. I can provide pointers if needed.

romainrey:
Hi Julien, Thanks for your quick answer, as usual!

Also another use case in addition to the one I was describing is if you are planning a full week at the same time and the urgency of some job is different if they are done the Monday or the Friday. You could represent day and vehicle as a couple vehicle-day and then the priority would be specific for this pair, which would help solve the problem of the urgency of a job that changes.

On the input data, yes but you could say that now for a job priority has to be a dict (vehicle1:priority1, ...) or an array if you assume knowing the order of the vehicles. But you could still let the possibility to only put an integer, and inside the function if it is only an integer then you just create the array yourself by replicating the integer. So that would not complexify the usage for those who don't want to use that option. For the others, I think that it is a very small cost for them for something that solve their problem.
[...]

PR #396, opened 2020-09-26
Priorities is a vector, having a value for each vehicle

romainrey:
## Issue
#393 
Please link to an existing issue (or open one before submitting a PR).
This allow to give vectors as priority or just an integer (as it is for the moment).
If the option vector is chosen then it should be vector of the size = number of vehicles

## Tasks
Need to validate that the optim is still ok, looks to work on mine.

 - [ ] ...
 - [ ] review


jcoupey:
@romainrey thanks for working on this and sharing the result!

Before going further, there two things that require investigating.

1. For some reason, the PR diff seems oversized: looks like basically all files are touched with a `+19,157 −18,761` diff, even when there is no perceptible change. Wild guess: maybe your editor replaced the EOL character across the whole codebase?
2. The CI build [failed](https://travis-ci.org/github/VROOM-Project/vroom/builds/730562965) with a compilation warning and an execution error.

romainrey:
Hi, sorry yes you were write, EOL issue...
Also I corrected the code, I just forgot the case where priorities were not given.

Let me know what do you think.

The solution can probably be more optimal because:
- I am not a c++ pro, my day to day job is in Python & R
- Even tho I tried to read the code, I didn't read all the files obviously and so I don't necessarily have the perfect "global" vision of the program.

jcoupey:
@romainrey it looks like this does a good job implementing the [first sketch](https://github.com/VROOM-Project/vroom/issues/393#issuecomment-696777480). Do you have some feedback on whether this is providing better solutions for the initial use-case you described?

I also realized I may have been overly optimistic on the changes required to implement this thoroughly. The current changes will do a good job putting higher priority jobs first in some phases of the search, mostly the new `UnassignedExchange` operator and the `try_job_addition` phases. The problem is that the priority objective is still conflicting with the travel time objective for all other local search operators. In other words, you could have, say `try_job_addition` putting in some higher priority job for a given vehicle route, and then a simple `Relocate` move switching that job back to another route because it's cheaper in term of travel time, regardless of the priority loss.

Managing this problem would require to extend all local search operators involving pairs of routes to add a notion of priority gain. This would mean an even deeper change to the codebase. Then frankly I'm not 100% sure I would be OK to merge this because it would mean having to maintain important changes in the long run for a feature that has not been showing up as significant in my own use-cases.

matejpavlovic-inits:
Is this still possible to implement?
I have a situation where I need to plan routes for, let's say, the next 5 days. My idea is to assign a different vehicle for each day. However, I'm not entirely sure how to handle urgent jobs that weren't completed today and need to be prioritized for tomorrow and so forth.

jcoupey:
Priority are about whether or not tasks are included in the plan *at all*. Constraints like "need to be prioritized for tomorrow" rather hint at using time windows.

matejpavlovic-inits:
It would be great if i could add, if this job is done by vehicle A then reward (priority) is 10, but if it is done by vehicle B then reward is 5

jcoupey:
Then you may want to give this branch a try to see if it works for your use-case.

PR #426, opened 2021-01-05
Cleanup includes

### LINKED ISSUE #425: Includes cleanup
I came across this blog post https://www.fluentcpp.com/2021/01/01/include-what-you-use/ and it would be interesting to see how `include-what-you-use` performs on the codebase.

jcoupey:
## Issue

Fixes #425. Since this PR is likely to touch many files anyway, I'm also taking the opportunity to update the copyright date to 2021.

## Tasks

 - [x] Run `iwyu` on the codebase
 - [x] Try and apply suggested changes
 - [x] Update copyright from 2020 to 2021
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Reduce build time by refactoring includes (#425)
+
 ## [v1.8.0] - 2020-09-29
 
 ### Added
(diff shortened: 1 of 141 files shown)


jcoupey:
The CI builds are passing so I think that's about it. We can be pretty confident on not breaking anything when using gcc or clang on Ubuntu. It would be nice to have some feedback on other systems though, as we've seen some unexpected build errors related to includes in the past. Yes I do mean MacOS.

PR #427, opened 2021-01-05
 Refactor/try job additions

### LINKED ISSUE #392: Improve try_job_additions
The `LocalSearch::try_job_additions` function is used to try adding unassigned jobs during the local search phase. It happens [after any route change](https://github.com/VROOM-Project/vroom/blob/43c09d0dd9d99fe3441bd12a4cbbc2ffdaa202ee/src/algorithms/local_search/local_search.cpp#L1231-L1233) that may give room to new insertions, or [after loosening the solution](https://github.com/VROOM-Project/vroom/blob/43c09d0dd9d99fe3441bd12a4cbbc2ffdaa202ee/src/algorithms/local_search/local_search.cpp#L1372), which we do once in a while when reaching a local minima.

Setting aside a couple tricks (handling shipments, shortcuts based on priorities), the current implementation sketch is as follows: 

1. Go through all unassigned jobs, then through all routes, then all potential insertion ranks to find out the best overall (valid) option in term of cost increase.
2. Apply the best option.
3. Loop back to 1. while there are available insertion options.

We do need to compute the information of the best cost for all job/route pair with associated insertion rank, but we are basically running the whole process at step 1. for each new choice, while most of the previous information are still valid and could be stored. So when inserting job `j` in route for vehicle `v`, we should keep all data for other jobs/vehicles, but only recompute choices across remaining unassigned jobs and vehicle `v`.

Of course this would only make a small difference for instances with no unassigned jobs. However, the time spent in `try_job_additions` can amount to nearly 20% of the overall running time in a situation with many unassigned jobs, like in [this flame graph](https://covid19.framadrop.org/lufi/r/wMMsPAvXnr#6FivkGgb9BBoxYRZaoiB1/ejvZ7dotO+Ht7IcaMOAv8=).

krypt-n:
## Issue

Closes #392

## Tasks

 -  ~Apply some more micro optimisations I have an a local branch somewhere~
 - ~Look into allocations in `compute_best_insertion_pd`~
 - [x] Benchmark properly
 - [x] review

This is my current progress on #392, the results so far seem promising. For comparison, the li_lim_100 benchmarks:

Baseline, current master@https://github.com/VROOM-Project/vroom/commit/24a6dd54175f40f2fb58be1c2ff76dd951a957f8:
```csv
,Gaps,Computing times
Min,-20.04,148
First decile,-0.0,301
Lower quartile,0.0,557
Median,0.0,685
Upper quartile,1.25,913
Ninth decile,4.19,986
Max,9.86,1236
```

This branch:
```csv
,Gaps,Computing times
Min,-20.04,114
First decile,-0.0,184
Lower quartile,0.0,231
Median,0.0,307
Upper quartile,1.25,397
Ninth decile,4.19,478
Max,9.86,601
```

Additionally I verified that the solutions for a range of different selected benchmarks are exactly the same. This makes me quite confident that my refactoring is correct. I'm going to continue working on this some time in the next couple of days.

### DIFF
--- a/src/algorithms/local_search/local_search.cpp
+++ b/src/algorithms/local_search/local_search.cpp
@@ -123,7 +123,7 @@ struct RouteInsertion {
 const RouteInsertion empty_insert = {std::numeric_limits<Gain>::max(), 0, 0, 0};
 
 template <class Route>
-RouteInsertion compute_best_insertion(const Input& input, const Index j, Index v, const Route& route) {
+RouteInsertion compute_best_insertion_single(const Input& input, const Index j, Index v, const Route& route) {
   const auto& matrix = input.get_matrix();
   RouteInsertion result = empty_insert;
   const auto& current_job = input.jobs[j];
@@ -266,6 +266,20 @@ RouteInsertion compute_best_insertion_pd(const Input& input, const Index j, Inde
   return result;
 }
 
+template <class Route>
+RouteInsertion compute_best_insertion(const Input& input, const Index j, Index v, const Route& route) {
+
+  const auto& current_job = input.jobs[j];
+  if (current_job.type == JOB_TYPE::SINGLE) {
+      return compute_best_insertion_single(input, j, v, route);
+  } else if (current_job.type == JOB_TYPE::PICKUP) {
+      return compute_best_insertion_pd(input, j, v, route);
+  } else {
+    return empty_insert;
+  }
+}
+
+
 template <class Route,
           class UnassignedExchange,
           class Exchange,
@@ -325,18 +339,9 @@ void LocalSearch<Route,
       }
       best_insert.assign(routes.size(), empty_insert);
 
-      if (current_job.type == JOB_TYPE::SINGLE) {
-        for (std::size_t i = 0; i < routes.size(); ++i) {
-          const auto v = routes[i];
-          best_insert[i] = compute_best_insertion(_input, j, v, _sol[v]);
-        }
-      }
-
-      if (current_job.type == JOB_TYPE::PICKUP) {
-        for (std::size_t i = 0; i < routes.size(); ++i) {
-          const auto v = routes[i];
-          best_insert[i] = compute_best_insertion_pd(_input, j, v, _sol[v]);
-        }
+      for (std::size_t i = 0; i < routes.size(); ++i) {
+        const auto v = routes[i];
+        best_insert[i] = compute_best_insertion(_input, j, v, _sol[v]);
       }
 
       auto smallest = std::numeric_limits<Gain>::max();
--- a/src/algorithms/local_search/local_search.cpp
+++ b/src/algorithms/local_search/local_search.cpp
@@ -113,6 +113,159 @@ LocalSearch<Route,
     });
 }
 
+
+struct RouteInsertion {
+  Gain cost;
+  Index single_rank;
+  Index pickup_rank;
+  Index delivery_rank;
+};
+const RouteInsertion empty_insert = {std::numeric_limits<Gain>::max(), 0, 0, 0};
+
+template <class Route>
+RouteInsertion compute_best_insertion(const Input& input, const Index j, Index v, const Route& route) {
+  const auto& matrix = input.get_matrix();
+  RouteInsertion result = empty_insert;
+  const auto& current_job = input.jobs[j];
+  const auto& v_target = input.vehicles[v];
+
+  if (!input.vehicle_ok_with_job(v, j)) {
+      return result;
+  }
+
+  for (Index rank = 0; rank <= route.size(); ++rank) {
+      Gain current_cost = utils::addition_cost(input,
+              matrix,
+              j,
+              v_target,
+              route.route,
+              rank);
+      if (current_cost < result.cost and
+              route.is_valid_addition_for_capacity(input,
+                  current_job.pickup,
+                  current_job.delivery,
+                  rank) and
+              route.is_valid_addition_for_tw(input, j, rank)) {
+          result = {current_cost, rank, 0, 0};
+      }
+  }
+  return result;
+}
+
+template <class Route>
+RouteInsertion compute_best_insertion_pd(const Input& input, const Index j, Index v, const Route& route) {
+  const auto& matrix = input.get_matrix();
+  RouteInsertion result = empty_insert;
+  const auto& current_job = input.jobs[j];
+  const auto& v_target = input.vehicles[v];
+
+  if (!input.vehicle_ok_with_job(v, j)) {
+      return result;
+  }
+
+  // Pre-compute cost of addition for matching delivery.
+  std::vector<Gain> d_adds(route.size() + 1);
+  std::vector<unsigned char> valid_delivery_insertions(
+            route.size() + 1);
+
+  for (unsigned d_rank = 0; d_rank <= route.size(); ++d_rank) {
+      d_adds[d_rank] = utils::addition_cost(input,
+              matrix,
+              j + 1,
+              v_target,
+              route.route,
+              d_rank);
+      valid_delivery_insertions[d_rank] =
+          route.is_valid_addition_for_tw(input, j + 1, d_rank);
+  }
+
+  for (Index pickup_r = 0; pickup_r <= route.size(); ++pickup_r) {
+      Gain p_add = utils::addition_cost(input,
+              matrix,
+              j,
+              v_target,
+              route.route,
+              pickup_r);
+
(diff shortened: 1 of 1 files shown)


jcoupey:
Sounds great! From what I can tell from the current commits, this already covers the scope of #392.

I'd like to run some benchmarking on my side too, but maybe I should wait as you listed more todo items? Or if the other items are not directly related to this refactoring, maybe we can move on with the current change and have them in a separate PR?

krypt-n:
You're right, #392 is basically done. Getting rid of the allocations in `compute_best_insertion_pd` seems to me to be a bit more invasive. I'd like to follow up on that though, since that and `PDShift` (similar algorithm) are responsible for around 50% of the calls to malloc in my benchmark.

I'd still like to add a few comments and format the code in a future commit, but feel free to benchmark and review this as is.
The commits should be reviewable individually. The first four only move stuff around and should not change behaviour in any way, the last one reduces the number of insertion calculations.

jcoupey:
Yes, I took a look at the PR with incremental diffs. Thanks for cutting out the changes in a few commits like you did, makes it really easy to follow on the overall logic.

I'll just fire a few solving rounds on the usual benchmarks which should provide a ballpark figure on the speedup anyway. Take the time you need to go on and I'll simply review the current state for a couple comments.

src/algorithms/local_search/local_search.cpp
```diff
@@ -113,6 +113,173 @@ LocalSearch<Route,
...
 }
 
+
+struct RouteInsertion {
+  Gain cost;
+  Index single_rank;
+  Index pickup_rank;
+  Index delivery_rank;
+};
+const RouteInsertion empty_insert = {std::numeric_limits<Gain>::max(), 0, 0, 0};
```
jcoupey, line 123: Is it possible to `constexpr` this?

src/algorithms/local_search/local_search.cpp
```diff
@@ -113,6 +113,173 @@ LocalSearch<Route,
...
+  const auto& matrix = input.get_matrix();
+  RouteInsertion result = empty_insert;
+  const auto& current_job = input.jobs[j];
+  const auto& v_target = input.vehicles[v];
+
+  if (!input.vehicle_ok_with_job(v, j)) {
+      return result;
+  }
+
+  for (Index rank = 0; rank <= route.size(); ++rank) {
```
jcoupey, line 136: In order to avoid an early `return` branch and only return at the end of the function, I'd be in favor of putting the `for` loop into an `if` with the opposite condition.

src/algorithms/local_search/local_search.cpp
```diff
@@ -113,6 +113,173 @@ LocalSearch<Route,
...
+template <class Route>
+RouteInsertion compute_best_insertion(const Input& input, const Index j, Index v, const Route& route) {
+
+  const auto& current_job = input.jobs[j];
+  if (current_job.type == JOB_TYPE::SINGLE) {
+      return compute_best_insertion_single(input, j, v, route);
+  } else if (current_job.type == JOB_TYPE::PICKUP) {
+      return compute_best_insertion_pd(input, j, v, route);
+  } else {
+    return empty_insert;
```
jcoupey, line 278: If I'm correct, this is never used, except if the caller incorrectly filters out `JOB_TYPE::DELIVERY`? If so it could make sense here to assert that the type is either `JOB_TYPE::SINGLE` or `JOB_TYPE::PICKUP`, then the `if`s could boil down to a simple ternary operator.

src/algorithms/local_search/local_search.cpp
```diff
@@ -147,20 +314,33 @@ void LocalSearch<Route,
...
   bool job_added;
-  std::vector<Gain> best_costs;
-  std::vector<Index> best_ranks;
-  std::vector<Index> best_pickup_ranks;
-  std::vector<Index> best_delivery_ranks;
+    
+  std::vector<std::vector<RouteInsertion>> route_job_insertions;
+
+  for (std::size_t i = 0; i < routes.size(); ++i) {
+      //TODO: tweak such that the vectors don't include all jobs
```
jcoupey, line 323: Have you considered mapping `RouteInsertion` objects with unassigned ranks as key (`j` value in the next loop)? That would make `route_job_insertions` a `std::vector<std::unordered_map<Index, RouteInsertion>>`.

If this is less efficient than a plain `std::vector`, I think we can leave with the extra non-used ranks for assigned jobs.

jcoupey (commented):
Dropped a couple comments. Also `./scripts/format.sh` generates a lot of changes, mostly indenting stuff.

jcoupey:
The following table shows the difference between 24a6dd5 (`master`) and 3f2118f (this PR) for various benchmarks. Average computing time variation (in %) is reported across several exploration levels. Note that for small `x` values the comparisons are not always meaningful as the computing times are sometimes just a few milliseconds (Solomon and Li & Lim 100). All solutions are strictly identical.

| Benchmark \ `x=` | 0 | 1 | 2 | 3 | 4 | 5 |
|------- | ---| ---| ---| ---| ---| ---|
| CVRP  | 0.7 | -11.2 | -13.0 | -17.5 | -21.8 | -26.5 |
| Solomon  | -1.8 | -1.3 | -4.2 | -3.6 | -6.8 | -9.4 |
| Homberger 200 | 2.7 | 1.0 | -1.3 | -5.0 | -9.2 | -12.0 |
| Li & lim 100 | -18.2 | -5.6 | -29.2 | -41.2 | -48.0 | -53.4 |
| Li & lim 200 | -3.3 | -8.7 | -30.1 | -46.8 | -54.3 | -59.8 |

When I spotted the problem and wrote the ticket, I though we'd get nice improvements but I did not suspect this order of magnitude. Especially for instances where there are actually no unassigned jobs (all except for a few CVRP instances here).

So the good news is: this already has a significant impact on computing times for the use of `try_job_additions` as part of the normal local search process, which happens when we loosen the routes and re-insert jobs to move on to another exploration level. The higher the exploration level, the more intensively we use `try_job_additions`, which explains the changes across `x` values.

Instances with shipments (Li & Lim classes) see a much bigger gain, which is in line with the fact that the insertion checks have a higher complexity than for regular jobs, so skipping the unnecessary checks is even more profitable.

Todo: I'll benchmark this at some point on instances with more unassigned jobs and we should see an even bigger impact.

jcoupey:
> Instances with shipments (Li & Lim classes) see a much bigger gain, which is in line with the fact that the insertion checks have a higher complexity

@krypt-n all the more interested to see the improvements you have in store for `compute_best_insertion_pd`.

src/algorithms/local_search/local_search.cpp
krypt-n: I tried dropping in a `std::unorderd_map` as you suggested but it increased the running time of the Li&lim instances I tried by 20-30%, that's why I removed the comment and left the vector as is.

krypt-n:
I briefly looked into the performance of the homberger instances with exploration level 0, but could not find a reliable slowdown on my machine on any instance. Do you know of any instance that performs worse on with this PR, or is the 2.7% just noise?

Other than that, I formatted this PR and applied the changes you suggested during the review

jcoupey (approved):

jcoupey:
Great! Looks like this is ready to merge.

My mistake on the Homberger 200 instances: I reported the difference for the computing time on the last instance only (364 ms vs 374 ms). The average is 228 ms for both runs. I just updated the table above.

Last question: you crossed the first two items in the list, does it mean you did not look into them for this PR, or that you did not find any interesting improvements?

krypt-n:
Ah great. I wasn't sure about that since the running time varies quite a bit from run to run and I didn't do enough benchmark runs to determine if the distribution changed.

I crossed them from the list because I don't feel they need to be part of this PR. I will most likely open new PRs for them, if I get any noteworthy results.

Yep, looks ready to merge for me as well.

jcoupey:
For completeness, I've generated random problems with time windows and 600 tasks using only 10 vehicles so more or less half the tasks end up unassigned.

| Instance \ `x=` | 0 | 1 | 2 | 3 | 4 | 5 |
|------- | ---| ---| ---| ---| ---| ---|
| 600 jobs |  1.4 | -1.6 | -4.9 | -6.1 | -2.5 | -8.5 |
| 300 shipments  |  -1.6 | -42.7 | -65.4 | -67.6 | -71.9 | -74.6 |

Of course that's only a couple instances but modulo computing time variations across runs, this is pretty much in line with previous tests and expectations.

jcoupey:
@krypt-n about the other potential optimizations you first listed, feel free to open a dedicated ticket if it makes sense, if only as a way to track the idea.

PR #449, opened 2021-02-02
VS 2019 compatibility

### LINKED ISSUE #448: Make lib and main win compatible
Currently there's lots of errors when compiling vroom on windows with visual studio.

nilsnolde:
fixes #448 

Note, I only tested building the static vroom library and the vroom main executable with these changes. I confirmed with [example_2](https://github.com/VROOM-Project/vroom/blob/master/docs/example_2.json) that it works in an offline scenario. I didn't have the time (yet) to try anything else.

Also, no idea how to manually build this with VS toolchain (and zero interest wanting to know;)), I actually used CMake. Which I'll contribute separately (gotta give that a test on linux first). 

In any case, the changes are a few missing includes and some ifdefs, nothing intrusive. Oh, and `getopt` is not available for MSVC, so I needed to borrow one from mingw64.

## Tasks

 - [ ] review


.gitignore
```diff
@@ -2,3 +2,4 @@
 *.d
 bin/
 lib/
+build/
```
nilsnolde, line 5: cmake prefers to have not in-source builds. would be great to keep it if you don't mind

src/utils/getopt_win.h
```diff
@@ -0,0 +1,653 @@
...
+ * by Dieter Baron and Thomas Klausner.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
```
nilsnolde, line 44: maybe exclude this file in the `makefile`? then you don't have to worry about it when you're shipping it to clients in non-VS situations

nilsnolde:
And don't sweat it if you don't want to include VS support.. This is really easy to maintain and patch in a fork. Same with CMake support..

jcoupey:
Thanks for submitting a PR! I have no real interest for win builds myself and I guess few users have based on the low number of complaints we've had so far. ;-)

On the other hand, if this only means adjusting includes and some compilation-related stuff, I'm fine with including the changes.

What bothers me most is the need to have an unrelated `getopt` implementation in the codebase. Could we at least have it live in the `include` directory alongside the `rapidjson` dependency?

nilsnolde:
> have it live in the include directory 

Sure, better idea anyways.

nilsnolde:
Also tested ORS on windows quickly btw to make sure the HTTP stuff is working as well.

src/utils/getopt_win.h
nilsnolde: moved it to include dir

.gitignore
nilsnolde: feel free to reject..

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 ### Fixed
 
 - Rapidjson assert on invalid syntax for first vehicle (#418)
+- Compatibility with Visual Studio 2019 (#448)
```
jcoupey, line 25: Now that we have a `release/1.9` branch, this should appear in a new `Unreleased` entry. I'll push the usual technical commit on `master` to allow this.

src/structures/typedefs.h
```diff
@@ -16,6 +16,20 @@ All rights reserved (see LICENSE).
...
+#ifdef _MSC_VER
+  // include support for "and"/"or"
+  #include <iso646.h>
+  // some include defines MAX_PRIORITY..
+  #ifdef MAX_PRIORITY
+    #undef MAX_PRIORITY
+  #endif
+  // ERROR is already defined in WIN
+  #ifdef ERROR
+    #undef ERROR
```
jcoupey, line 30: I don't get why this is required, since our own `MAX_PRIORITY` and `ERROR` are defined within the `vroom` namespace?

nilsnolde: Actually I think I can get rid of the ERROR thing at least by providing the `NOGDI` directive during compilation, WinGDI.h defines ERROR as macro. Why that interferes with scoped names... :shrug: No idea what defines `MAX_PRIORITY`, must be some hidden macro as well..

CHANGELOG.md
nilsnolde: just realized I forgot smth here.. I can do it as well then.

src/main.cpp
```diff
@@ -10,7 +10,11 @@ All rights reserved (see LICENSE).
 #include <fstream>
 #include <iostream>
 #include <sstream>
-#include <unistd.h>
+#ifndef _WIN32
+  #include <unistd.h>
+#else
+  #include "utils/getopt_win.h"
```
nilsnolde, line 16: urgh.. I did change the include here without utils.. was sure I committed that! will rectify

nilsnolde:
Let me do the following once I'm back on win:

- adapt changelog
- fix the include for the new location of `getopo_win.h`
- experiment with some directives to not include Win headers defining some crap macros

nilsnolde:
alright, impact on the codebase minimized:) good to go from my side

For future reference: this won't compile with MSVC unless you include the `NOGDI` preprocessor directive, which makes sure WinGDI.h is not pulled in causing all sorts of trouble. For good measure, include `WIN32_LEAN_AND_MEAN`, bringing down the compile time by a minute or two..

CHANGELOG.md
nilsnolde: done

src/structures/typedefs.h
nilsnolde: NOGDI was fixing both!

src/main.cpp
nilsnolde: done

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Compatibility with Visual Studio 2019 (#448)
+
```
jcoupey, line 8: This is the source of the current conflict, as I've pushed an update commit in the meantime.

jcoupey (commented):
Except for the changelog conflict, this is good to merge.

CHANGELOG.md
nilsnolde: oops

PR #459, opened 2021-02-21
Implement reading input from stdin

### LINKED ISSUE #457: read json problem from stdin
```bash
cat myproblem.json | vroom -i - 
```

this not work, actually vroom accept only input as string in last parameter?? or a filename with -i param?

I would need to pass input json in stdin in my nodejs spawn

krypt-n:
## Issue

Fixes #457

## Tasks

 - [x] Update `CHANGELOG.md` 
 - [ ] review

## Description

With these changes, vroom follows the following strategy for getting its input:
1. If there are spare arguments, read the first spare argument as input
2. If the `-i FILE` option was used, read `FILE` as input
3. Otherwise read `stdin` as input

Previously vroom would display its help string in case 3. This degrades the experience for new users somewhat, since executing `vroom` without any arguments won't appear to do anything until the user closes `stdin` with `^D`, which then results in a cryptic json parsing exception.

Alternatives to consider:
* Only read from `stdin` as input if `stdin` is not a tty (unsure how portable `isatty` is)
* Display help string if the input read from `stdin` has length 0 or is not valid json

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 
 - Formatting script updated to use version 10 of clang-format (#452)
+- vroom will now read json input from stdin if no other input is specified (#457)
 
 ### Fixed
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -32,6 +32,7 @@ void display_usage() {
   usage += "Version: " + vroom::get_version() + "\n";
   usage += "Usage:\n\tvroom [OPTION]... \"INPUT\"";
   usage += "\n\tvroom [OPTION]... -i FILE\n";
+  usage += "\tvroom [OPTION]...\n";
   usage += "Options:\n";
   usage += "\t-a PROFILE:HOST (=" + vroom::DEFAULT_PROFILE +
            ":0.0.0.0)\t routing server\n";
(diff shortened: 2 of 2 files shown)


jcoupey:
Thanks for submitting a PR!

> Display help string if the input read from stdin has length 0 or is not valid json

In that case the error will be caught down the line upon parsing so I think we can keep things that way. It would be consistent with the current behavior, both `vroom ''` and `vroom -i missing_file.json` currently error with `{"code":2,"error":"The document is empty. (offset: 0)"}`.

> Only read from stdin as input if stdin is not a tty 

I'm not sure to get the implications here but I'm not very familiar with this kind of stuff. In what situation would this be an improvement?

krypt-n:
> I'm not sure to get the implications here but I'm not very familiar with this kind of stuff. In what situation would this be an improvement?

Running vroom from a terminal without arguments would still print the help message (potentially helping new users) and running `cat problem.json | vroom` would work as expected. The downside is that there is no way to do this via the C++ standard library and we would have to use `isatty` for unix-like systems and something else for windows.

jcoupey:
I think we can live with `vroom` (without arguments) not displaying the help message, after all it's a case of misuse and there is an appropriate error message. Also we provide the `-h` flag, which is the way most command-line users would expect to get some help.

krypt-n:
Alright, I added this change to the changelog. This is ready to merge, I think

PR #461, opened 2021-02-24
Use Valhalla for routing

### LINKED ISSUE #306: Add Valhalla as routing provider
I think Valhalla, despite its relatively poor performance compared to ORS/OSRM, would be a nice addition to Vroom's routing providers. It's really unique in terms of run-time weights & costs, no other router can match that, except for BRouter. Anyways, with Mapbox ending OSRM support in a few weeks time and fully concentrating on Valhalla, it might also be a step towards future.

Of course it's possible to add Valhalla already via custom matrices. 

Actually I think I'd like to tackle this myself, if no one else jumps for it in the next weeks. Recently started some C++ dev and this seems rather easy.

jcoupey:
## Issue

Fixes #306

## Tasks

 - [x] Draft a `ValhallaWrapper` class as a `HttpWrapper` base class
 - [x] Make `HttpWrapper` more generic to handle stuff that currently works for both OSRM and ORS but not with Valhalla
 - [x] Implement matrix queries to Valhalla
 - [x] Implement route queries to Valhalla
 - [x] Update `CHANGELOG.md`
 - [ ] review


jcoupey:
I've given this a good start but I'm running into a problem with the current way the response reading goes:

https://github.com/VROOM-Project/vroom/blob/313e1f931b8a6884d1c21280edb5b3cae732cfb5/src/routing/http_wrapper.cpp#L51-L62

Normally, when all of the response has been received `read_some` will update the value of `error` to match `asio::error::eof` and we're out. With the instance of Valhalla I'm trying this on, it looks like the end of reception is never properly sent, so `read_some` hangs forever once `response` is fully populated. This is puzzling as firing the same requests with `curl` or `wget` works like a charm.

I have no idea if this is related to the server-side internals of Valhalla (prime-server?) or if this is even reproducible on other Valhalla instances. Until further digging I'll probably implement a hack around this based on the `Content-Length` header.

nilsnolde:
Hm, weird.. Never had that problem in any other client library (and never saw a Valhalla issue complaining about it), so my assumption is it's probably the way you do the request (although it seems to follow the asio tutorial..). 

Without having a closer look it's hard to tell for me, I also never build a client app in C++. Maybe @kevinkreiser has a clue off the top of his head, he's one of the main people behind valhalla and mostly built prime_server.

jcoupey:
FWIW the way the query is built is borrowed from what currently works for both OSRM and ORS:

https://github.com/VROOM-Project/vroom/blob/6313ca7dfc65a8b801d5aca85c6197061443e1a6/src/routing/valhalla_wrapper.cpp#L25-L48

Maybe some more header magic is required...

krypt-n:
I am not entirely sure if I found the correct code that is responsible for this but this line https://github.com/kevinkreiser/prime_server/blob/master/src/http_protocol.cpp#L250
seems to indicate that prime_server only respects `Connection: Close`, not `Connection: close` (case sensitive).

I tested it locally and `Connection: Close` seems to work

jcoupey:
@krypt-n thanks a lot for investigating and spotting the problem. (Sigh) I did waste a couple hours debugging and working around this...

jcoupey:
This is now pretty much in a working state. I had to take the trouble to merge route legs provided by Valhalla and included https://github.com/vahancho/polylineencoder as a dependency for this purpose.

kevinkreiser:
@jcoupey and @krypt-n sorry for the annoyance on this :frowning_face: i can make these case-less compares. ive always wanted to dump the http parsing in favor of proven http parsing library but i havent found the time.

marko-bogdanovic:
Hi I'm a beginner in this. I started Valhalla locally on the computer, but now I'm interested in how to do the integration with Vroom and show it all on a map using Leaflet. Any help would be most welcome, thanks in advance.

nilsnolde:
Give me a few minutes, see https://github.com/VROOM-Project/vroom-express/issues/77.

I'll add the defaults real quick in a PR, then you can take it from there.

You'll need the full stack, i.e. also the repo above, vroom-express. The easiest is a docker-compose setup, where you add [vroom docker](https://github.com/VROOM-Project/vroom-docker) (contains vroom and vroom-express) and e.g. [our valhalla image](https://github.com/gis-ops/docker-valhalla). Then you can call vroom-express via HTTP which returns the JSON response. You'll need to parse the JSON in JS and build your leaflet objects manually (i.e. the response is not a GeoJSON or similar), Julien's example repo probably helps: https://github.com/VROOM-Project/vroom-frontend.

nilsnolde:
Ok, here's the right config.yml: https://github.com/VROOM-Project/vroom-express/pull/78.

nilsnolde:
Also, be aware you can't (yet) set custom matrix and/or routing parameters, though we do hope to support that very soon. If you always need the same request parameters you can hardcode them in vroom though and re-compile the source.

marko-bogdanovic:
My config.yml file now looks like [this](https://github.com/VROOM-Project/vroom/files/6303960/config.txt). But now when I run npm start, I get the following error "Error: listen EACCES: permission denied 0.0.0.0:3000". How to change host 0.0.0.0 with my host ?

nilsnolde:
You're not using it right somehow.. Sorry, but I have no idea which steps you took to get there (apparently not the docker route, but rather host-native installation). Best to think about it again, go through the documentation and either open a new issue in the appropriate repository or rather stackoverflow, as there's nothing wrong with vroom(-express) on a technical side.

marko-bogdanovic:
Now I have changed port and launched vroom-express, but I get an error **"/ bin / sh: 1: vroom: not found"**. Sorry for too many questions I am an absolute beginner in this.

jcoupey:
You'll have to check your `vroom-express` configuration and use the relevant `vroom` path for you own setup.

I'm locking this conversation as it initially only had to do with integrating Valhalla in the core solving engine. @marko-bogdanovic please make sure to post issues in the relevant repository.

PR #464, opened 2021-03-01
Change pd_shift loop range

### LINKED ISSUE #462: Question regarding pd_shift
Hi, this is not a real issue, just a thing I noticed in the code.

I'm currently looking into the running time of instances that use PDShift and compute_best_insertion_pd heavily.
Both contain an algorithm to find a cheap insertion of a shipment into a route. I merged the two algorithms but noticed that this changes the resulting routes.
After some debugging I noticed that the only relevant difference between the algorithm in pd_shift.cpp and the one in local_search.cpp is the range of the possible pickup locations. My question thus is:

Is it intended that PDShift does not consider the end of a route for the pickup? Or am I missing something?

This change is not a clear improvement in solution quality in the benchmarks I checked, some solutions improve, some get worse. This is why I am hesitant to apply this change.

```diff
diff --git a/src/problems/vrptw/operators/pd_shift.cpp b/src/problems/vrptw/operators/pd_shift.cpp
index 39f2a7b..0dcefd2 100644
--- a/src/problems/vrptw/operators/pd_shift.cpp
+++ b/src/problems/vrptw/operators/pd_shift.cpp
@@ -80,7 +80,7 @@ void PDShift::compute_gain() {
       target.is_valid_addition_for_tw(_input, s_route[_s_d_rank], t_d_rank);
   }

-  for (unsigned t_p_rank = 0; t_p_rank < t_route.size(); ++t_p_rank) {
+  for (unsigned t_p_rank = 0; t_p_rank <= t_route.size(); ++t_p_rank) {
     Gain t_p_gain = -utils::addition_cost(_input,
                                           m,
                                           s_route[_s_p_rank],
```

krypt-n:
## Issue

Fixes #462

## Tasks

 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Compatibility with Visual Studio 2019 (#448)
+- the pd_shift operation can now insert a shipment at the end of a route (#462)
 
 ## [v1.9.0-rc.1] - 2021-01-14
 
--- a/src/problems/vrptw/operators/pd_shift.cpp
+++ b/src/problems/vrptw/operators/pd_shift.cpp
@@ -80,7 +80,7 @@ void PDShift::compute_gain() {
       target.is_valid_addition_for_tw(_input, s_route[_s_d_rank], t_d_rank);
   }
 
-  for (unsigned t_p_rank = 0; t_p_rank < t_route.size(); ++t_p_rank) {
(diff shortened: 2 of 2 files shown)


jcoupey (approved):
Thanks for the PR, we just need a changelog entry in the "fixed" section.

krypt-n:
Done, I added a changelog entry

PR #491, opened 2021-04-21
Optimize cost_wrapper

### LINKED ISSUE #490: Check cost computing overhead
The ability to use multiple profiles introduced in #450 changed the way we now access cost values. Instead of a single matrix lookup previously, we now  have a ponderation with a vehicle-dependent speed factor happening in `Vehicle::Cost`:

https://github.com/VROOM-Project/vroom/blob/1e32124bbfdf131364acbd52109628707aa0c284/src/structures/vroom/vehicle.h#L67-L71

This is a price we have to pay for the sake of a much bigger flexibility in cost evaluation. The downside is that this also means a computing time increase for single-profile instances even with unused vehicle `speed_factor` (see https://github.com/VROOM-Project/vroom/pull/450#issuecomment-816561991).

I've tried to make this as efficient as possible, in particular by providing the `Vehicle::Cost` implementation in header file to allow the compiler to perform some optimizations in other units. I wonder if there is more to it and if we could improve the structures or the layout to further reduce the increase. I guess this would call for some profiling to get a clearer view of what are the most expensive bits here.

krypt-n:
## Issue

Closes #490 

## Tasks

 - [x] review

## Description
hommberger_200 VRPTW benchmark master at 1e32124bbfdf131364acbd52109628707aa0c284

```
,Gaps,Computing times
Min,-14.89,194
First decile,-4.56,386
Lower quartile,-0.0,482
Median,2.03,684
Upper quartile,6.67,1690
Ninth decile,8.88,2070
Max,11.53,2491
```

The first commit https://github.com/VROOM-Project/vroom/commit/3a0e9d9e2dfcb4e3d596a6f46b9046417e132e6e turns `Matrix`from a vector of vectors into a flat vector. This removes a layer of indirection and could perhaps improve memory locality.

```
,Gaps,Computing times
Min,-14.89,183
First decile,-4.56,377
Lower quartile,-0.0,478
Median,2.03,671
Upper quartile,6.67,1660
Ninth decile,8.88,2014
Max,11.53,2461
```

The second commit https://github.com/VROOM-Project/vroom/commit/3a0e9d9e2dfcb4e3d596a6f46b9046417e132e6e just moves some code around and should not have an impact.

```
,Gaps,Computing times
Min,-14.89,186
First decile,-4.56,375
Lower quartile,-0.0,480
Median,2.03,672
Upper quartile,6.67,1654
Ninth decile,8.88,2025
Max,11.53,2465
```

The third commit https://github.com/VROOM-Project/vroom/commit/9cba73ef6ad9540cf7552b879ef847020204fc9a removes another layer of indirection by allowing `cost_wrapper`to directly access the matrix array.

```
,Gaps,Computing times
Min,-14.89,181
First decile,-4.56,371
Lower quartile,-0.0,463
Median,2.03,664
Upper quartile,6.67,1623
Ninth decile,8.88,1994
Max,11.53,2407
```

My next idea was to get rid of the cast to double by turning the cost factor into a whole number. My first attempt was https://github.com/VROOM-Project/vroom/commit/510557e4c609e690dcb02d7ec60a5bf2c02890ba .

```
,Gaps,Computing times
Min,-14.89,162
First decile,-4.56,332
Lower quartile,-0.0,438
Median,2.03,646
Upper quartile,6.67,1445
Ninth decile,8.88,1872
Max,11.53,2369
```
However, it turned out that 128 as a factor did not lead to correct reported costs on the HVRP instances (the cost factors can not be exactly represented as multiples of 1/128). That's why I changed the factor to 100 and fixed an additional rounding problem in https://github.com/VROOM-Project/vroom/commit/eb253e2c141b75c71119488a86109704ac2a4b13 and https://github.com/VROOM-Project/vroom/commit/6b3e488d3e5cd714e2f082209895ecc92a8a332d . Leading to these final results:

```
,Gaps,Computing times
Min,-14.89,166
First decile,-4.56,343
Lower quartile,-0.0,455
Median,2.03,615
Upper quartile,6.67,1537
Ninth decile,8.88,1945
Max,11.53,2243
```
Note that these computing times tend to fluctuate quite a bit on my machine, but I think the general trend towards lower computing times is visible here.
The last change effectively restricts the duration factor to two digits after the decimal point. That should probably be documented somewhere if this PR is merged.

### DIFF
--- a/src/structures/vroom/cost_wrapper.h
+++ b/src/structures/vroom/cost_wrapper.h
@@ -16,7 +16,10 @@ All rights reserved (see LICENSE).
 namespace vroom {
 
 struct CostWrapper {
+  const static uint32_t DIVISOR = 128;
+
   const double durations_factor;
+  const uint32_t discrete_factor;
   const Matrix<Cost>* durations_matrix;
   
   std::size_t cost_matrix_size;
@@ -28,8 +31,7 @@ struct CostWrapper {
 
   Cost cost(Index i, Index j) const {
     Cost c = cost_data[i * cost_matrix_size + j];
-    return static_cast<Cost>(
-      durations_factor * static_cast<double>(c));
+    return (c * discrete_factor)/ DIVISOR;
   }
 
 };
--- a/src/structures/vroom/cost_wrapper.h
+++ b/src/structures/vroom/cost_wrapper.h
@@ -18,10 +18,7 @@ namespace vroom {
 struct CostWrapper {
   const static uint32_t DIVISOR = 100;
 
-  const double durations_factor;
   const uint32_t discrete_factor;
-  const Matrix<Cost>* durations_matrix;
-
   std::size_t cost_matrix_size;
   const Cost* cost_data;
 
--- a/src/structures/vroom/cost_wrapper.h
+++ b/src/structures/vroom/cost_wrapper.h
@@ -22,6 +22,13 @@ struct CostWrapper {
   CostWrapper(double speed_factor);
 
   void set_durations_matrix(const Matrix<Cost>* matrix);
+
+  Cost cost(Index i, Index j) const {
+    return static_cast<Cost>(
+      durations_factor *
+      static_cast<double>((*(durations_matrix))[i][j]));
+  }
+
 };
 
 } // namespace vroom
--- a/src/structures/vroom/cost_wrapper.h
+++ b/src/structures/vroom/cost_wrapper.h
@@ -18,15 +18,18 @@ namespace vroom {
 struct CostWrapper {
   const double durations_factor;
   const Matrix<Cost>* durations_matrix;
+  
+  std::size_t cost_matrix_size;
+  const Cost* cost_data;
 
   CostWrapper(double speed_factor);
 
   void set_durations_matrix(const Matrix<Cost>* matrix);
 
   Cost cost(Index i, Index j) const {
+    Cost c = cost_data[i * cost_matrix_size + j];
     return static_cast<Cost>(
-      durations_factor *
-      static_cast<double>((*(durations_matrix))[i][j]));
+      durations_factor * static_cast<double>(c));
   }
 
 };
--- a/src/structures/vroom/cost_wrapper.h
+++ b/src/structures/vroom/cost_wrapper.h
@@ -21,7 +21,7 @@ struct CostWrapper {
   const double durations_factor;
   const uint32_t discrete_factor;
   const Matrix<Cost>* durations_matrix;
-  
+
   std::size_t cost_matrix_size;
   const Cost* cost_data;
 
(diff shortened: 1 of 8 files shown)


krypt-n:
But it seems like I need to find a way to make these benchmarks produce more reliable numbers

jcoupey:
Thanks a lot for this input.

I quite like the `Matrix` refactor in the sense that it does not affect anything regarding usage, except in cost wrapper that makes use of the internal layout.

giphy

Keeping the cost scaling part in the integer world is definitely something I should have done in the first place! Restricting the duration factor to two digits after the decimal point is definitely not a problem. This does not _exactly_ apply to the `speed_factor` used in input as we use the inverse as `durations_factor`. But for the sake of simplicity I think we can we add that `speed_factor` "defaults to 1. with two digits after decimal point taken into account" in the docs.

src/structures/vroom/cost_wrapper.h
```diff
@@ -16,12 +16,23 @@ All rights reserved (see LICENSE).
 namespace vroom {
 
 struct CostWrapper {
+  const static uint32_t DIVISOR = 100;
+
   const double durations_factor;
+  const uint32_t discrete_factor;
   const Matrix<Cost>* durations_matrix;
```
jcoupey, line 23: It looks like we can get rid of the `durations_matrix` pointer now since we access `cost_data` directly.

jcoupey (commented):
Looking good with the single comment and docs update.

Regarding changelog, I don't think this should generate an entry as it's a change on a change, not wrt the previous release. But it would be useful to add #490 along with #394 in the first "Support for heterogeneous vehicle routing profiles" entry.

jcoupey:
On the topic of getting reliable numbers on benchmarks, the variation in computing times is probably more noticeable with small instances (homberger_200 vs homberger_1000). Maybe a good approach would be to sum computing times across a given number of runs on each instance, but that's another topic altogether.

For now we're interested in overall trends and I guess they'll be consistent across benchmark classes.

src/structures/vroom/cost_wrapper.h
krypt-n: Done, I removed `durations_factor`as well.

jcoupey:
I just merged this PR both to `master` and to `release/1.10` so it's now part of a new release candidate `v1.10.0-rc.2`. I'll run the usual benchmark stuff on my side and report.

jcoupey:
Running on the usual CVRP and VRPTW benchmarks across various exploration level shows a consistent improvement. The average computing time increase between `master` and #450 was ~17.6%. Now the increase between `v1.9.0` and `v1.10.0-rc.2` (including this PR) is down at 8.4% on average.

So the changes introduced here more than halved the impact on computing times for single-profile instances.

PR #513, opened 2021-05-14
Fix creation of matrix in example code

### LINKED ISSUE #510: vroom/libvroom_examples/libvroom.cpp compiling error
vroom/libvroom_examples/libvroom.cpp(218,42): error C2664: 'vroom::Matrix<vroom::Cost>::Matrix(size_t)': cannot convert argument 1 from 'initializer list' to 'size_t'

krypt-n:
## Issue

Fixes #510 

My recent PR broke the libvroom example, since I was not aware of the intializer_list constructor usage. 
Creating a hardcoded matrix is not a commin use case right now in my opinion, that's why this PR changes the example to use the available methods to create the matrix, instead of re-adding the initializer_list constructor.

## Tasks

 - [ ] review

### DIFF
--- a/libvroom_examples/libvroom.cpp
+++ b/libvroom_examples/libvroom.cpp
@@ -215,10 +215,25 @@ void run_example_with_custom_matrix() {
   vroom::Input problem_instance(amount_dimension);
 
   // Define custom matrix and bypass OSRM call.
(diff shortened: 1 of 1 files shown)


jcoupey:
@krypt-n thanks for the fix!

I filled #514 to remember adding a rule to the CI builds in order to spot this kind of mismatch early on.

Issue #455, opened 2021-02-08
Update CI build image to Ubuntu 20.04


jcoupey:
CI runs currently use Ubuntu bionic images, which means we are checking the build against compiler versions that are (much) older than the ones currently available out of the box in recent dev/production boxes. Switching to Ubuntu focal should enable updating:

- gcc from version 8 to 9;
- clang from version 6 to 10.

The way to do this is somehow related to the outcome of #436 but I'm still filling a dedicated ticket as we'll need it in any case at some point.

jcoupey:
Done in #517.

Issue #436, opened 2021-01-14
Migrate from Travis CI to GitHub Actions


jcoupey:
I've been lazily avoiding this for a while, but we're now past the official deadline for travis-ci.org shutdown. The support period has probably been somehow extended as the current build setup still works.

But things will eventually break so the move to travis-ci.com should be taken care of. My understanding of https://docs.travis-ci.com/user/migrate/open-source-repository-migration is that the move should not really change anything on our side, including still being able to use the service for free as an open-source project.

braktar:
travis-ci.com has recently changed its billing policy, which is extremely restrictive (1,000 seconds per month) and the quality of the service is variable.

Github actions still provides a comfortable free credit for for public repositories (2,000 minutes per month).

jcoupey:
Ah well, I was hoping to migrate without having to change too much stuff but then it looks like we'll have to make the switch.

Adjusting the current yaml config to migrate to GitHub Actions is probably not that big a deal. But since I have virtually no experience on the matter I'd gladly take views from anyone that has tried it already!

jcoupey:
My overall feeling about CI for this repo is that triggering a build for each push in a PR branch is basically a waste of resources. That's bad, even when using a free service. We don't have an instant deployment workflow justifying to do this, and in most situations we really just need a build check upon reviewing a PR or tagging releases, maybe merge to `master`.

I took a closer look on the GA workflows/event concepts and the variety of ways to trigger CI builds. This may make it possible to have a convenient way to trigger builds only when we want to. Ideas (not sure what would be the most convenient):

- manually;
- when a review is done;
- when a specific string appears in a PR comment;
- when a dedicated label is applied;
- using a specific git tag (mandatory for releases).

jcoupey:
In the meantime, I've made the switch to travis-ci.com. I removed older gcc and clang version checks so we're left we 2 checks out of 4 (gcc 8 and clang 6).

The 2 resulting builds (one from #451 and one from the `master` merge) seemed to consume only 90 of 10000 credits so we should be rather safe for the time being.

jcoupey:
Update: my understanding of the Travis plans is that "Free" comes with a **one time** allocation of 10 000 credits. On the other hand, there is a concept of "OSS only credits" that are monthly renewable and might be fine for this project.

[...]

PR #517, opened 2021-05-27
Ci/GitHub action migration

Orthae:
## Issue
Migrate from Travis CI to GitHub Actions #436
Update CI build image to Ubuntu 20.04 #455

## Tasks

 - [X] Moved CI to Github Actions
 - [X] Bumped Ubuntu to 20.04
 - [X] Bumped compilers to Clang 10/G++9
 - [X] Updated badges to new CI
 - [x] Review

### DIFF
--- a/.github/workflows/libosrm.yml
+++ b/.github/workflows/libosrm.yml
@@ -12,6 +12,8 @@ on:
      paths:
        - '.github/workflows/libosrm.yml'
        - '**libosrm_wrapper**'
+env:
+  osrm-tag: v5.25.0
 jobs:
   libosrm:
     strategy:
@@ -26,17 +28,15 @@ jobs:
       - name: Cache OSRM
         id: cache
         uses: actions/cache@v2
-        env:
-          cache-name: cache-node-modules
         with:
           path: ${{ github.workspace }}/osrm-backend
-          key: osrm-${{ matrix.cxx }}
+          key: osrm-${{ env.osrm-tag }}-${{ matrix.cxx }}
       - name: Checkout OSRM repository
         if: steps.cache.outputs.cache-hit != 'true'
         uses: actions/checkout@v2
         with:
           repository: Project-OSRM/osrm-backend
-          ref: v5.25.0
+          ref: ${{ env.osrm-tag }}
           path: osrm-backend
       - name : Install OSRM dependencies
         run: sudo apt-get install build-essential git cmake pkg-config libbz2-dev libstxxl-dev libstxxl1v5 libxml2-dev libzip-dev libboost-all-dev lua5.2 liblua5.2-dev libtbb-dev libluabind-dev libluabind0.9.1d1
--- a/.github/workflows/libosrm.yml
+++ b/.github/workflows/libosrm.yml
@@ -1,0 +1,45 @@
+name: Libosrm CI
+on:
+  pull_request:
+     branches:
+       - master
+     paths:
+       - '.github/workflows/libosrm.yml'
+       - '**libosrm_wrapper**'
+jobs:
+  libosrm:
+    strategy:
+      matrix: 
+        cxx: ['g++-9', 'clang++-10']
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout Vroom repository
+        uses: actions/checkout@v2
+      - name: Install Vroom dependencies
+        run: sudo apt-get install libasio-dev libglpk-dev jq
+      - name: Checkout OSRM repository
+        uses: actions/checkout@v2
+        with:
+          repository: Project-OSRM/osrm-backend
+          ref: v5.25.0
+          path: osrm-backend
+      - name : Install OSRM dependencies
+        run: sudo apt-get install build-essential git cmake pkg-config libbz2-dev libstxxl-dev libstxxl1v5 libxml2-dev libzip-dev libboost-all-dev lua5.2 liblua5.2-dev libtbb-dev libluabind-dev libluabind0.9.1d1
+        working-directory: osrm-backend
+      - name: Compile OSRM
+        run: |
+          mkdir build
+          cd build
+          cmake .. -DCMAKE_BUILD_TYPE=Release
+          cmake --build .
+        working-directory: osrm-backend
+      - name: Install OSRM
+        run: sudo cmake --build . --target install
+        working-directory: osrm-backend/build
+      - name: Compile Vroom
+        run: make
+        env:
+          CXX: ${{ matrix.cxx }}
+        working-directory: src
+      - name: Validate output
+        run: diff <(bin/vroom -i docs/example_2.json  | jq '.routes[].steps[]' --sort-keys) <(jq '.routes[].steps[]' --sort-keys docs/example_2_sol.json)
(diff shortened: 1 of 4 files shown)


nilsnolde:
Did you consider or try using OSRM docker image as the base image for the `libosrm` job and install vroom into it? It's fairly void of installed software, but maybe more efficient than building it from scratch every time incl deps?

Orthae:
No, I wasn't thinking of that. I could try using docker or try to utilize compiled binaries from the OSRM release page.

.github/workflows/libosrm.yml
```diff
@@ -0,0 +1,51 @@
...
+       - master
+     paths:
+       - '.github/workflows/libosrm.yml'
+       - '**libosrm_wrapper**'
+jobs:
+  libosrm:
+    strategy:
+      matrix: 
+        cxx: ['g++-9', 'clang++-10']
+    runs-on: ubuntu-20.04
```
nilsnolde, line 20: @Orthae GA makes running the workflow in your own image pretty easy, this should be enough:

```suggestion
    container: osrm/osrm-backend:v5.25.0
```

Then you should be able to simply remove the OSRM parts here and jump directly to compiling vroom in some directory.

Orthae: I checked it, and it seems that the entire pipeline has to be running inside the osrm container, and there will need for manually installing all dependencies. Also, the image is on Debian 9 which, gives access to g++6.

What about cache? Adding osrm to Github Actions cache after compiling, will help with subsequent runs when someone actively works on something related to libosrm.

nilsnolde: > gives access to g++6.

hmpf yeah that might be too old for vroom, not sure.. 

If you think cache is easy to implement here, sure.. I'm not very versed with caching builds, but could imagine that g++/clang++ matrix could create problems? Do they cache well? I usually use ccache, might be easier here as well?

nilsnolde: Disclaimer, I'm not in the vroom core team, so this is "just" a suggestion to save some CI hours. 

you're already restricting it to the few paths that are triggering this and the travis jobs didn't have any cache either. implementing cache might take more effort than it's worth maybe..

jcoupey:
@Orthae thanks a lot for working on that one and submitting a PR! Just a general comment before I find time to dive into the details. I don't think we should bother with changing the existing workflow with Docker right now because:

- it works™
- just porting the current travis behavior over to GA is already great;
- it's probably a good idea to validate the move without changing any other parameters;
- we can still change that later if it makes sense;
- the situation where we actually want to trigger a full `libosrm` + `vroom` build is rare (only when `libosrmwrapper.cpp` has been changed).

.github/workflows/libosrm.yml
Orthae: Added cache, subsequent build are much faster now.

Orthae:
@jcoupey No problem, I have added cache for libosrm anyway should be good for now.

.github/workflows/libosrm.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Install Vroom dependencies
+        run: sudo apt-get install libasio-dev libglpk-dev jq
+      - name: Cache OSRM
+        id: cache
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: ${{ github.workspace }}/osrm-backend
+          key: osrm-${{ matrix.cxx }}
```
nilsnolde, line 33: looks good @Orthae . just wondering if it should have the osrm version in the key, otherwise it won't have a cache miss when osrm is upgraded, no?

Orthae: You are correct, I will add OSRM tag to the cache key.

.github/workflows/libosrm.yml
```diff
@@ -0,0 +1,61 @@
+name: Libosrm CI
+on:
```
jcoupey, line 2: Would be nice to add a `tags` entry in order to trigger a build for all RC and final versions. The current Travis pattern is `/^v\d+\.\d+(\.\d+)?(-\S*)?$/` which is probably overly complicated, anything like `v*.*.*` would probably do.

.github/workflows/vroom.yml
```diff
@@ -0,0 +1,26 @@
+name: Vroom CI
+on:
+  push:
```
jcoupey, line 3: Same on tag build than in the other workflow.

.github/workflows/libosrm.yml
jcoupey: This cache thing is really neat. It will be much easier for us to check build against `libosrm` as it will only be built once (per version).

.github/workflows/libosrm.yml
```diff
@@ -0,0 +1,61 @@
...
+  push:
+     branches:
+       - master
+     paths:
+       - '.github/workflows/libosrm.yml'
+       - '**libosrm_wrapper**'
```
jcoupey, line 8: I would remove the `libosrm_wrapper` path restriction here, in order to systematically check build against `libosrm` for each new feature merge in `master` (that will be cheap enough thanks to caching). If we keep the restriction for `pull_request` we'll filter out most of the unnecessary  work anyway.

.github/workflows/libosrm.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Install OSRM
+        run: sudo cmake --build . --target install
+        working-directory: osrm-backend/build
+      - name: Compile Vroom
+        run: make
+        env:
+          CXX: ${{ matrix.cxx }}
+        working-directory: src
+      - name: Validate output
+        run: diff <(bin/vroom -i docs/example_2.json  | jq '.routes[].steps[]' --sort-keys) <(jq '.routes[].steps[]' --sort-keys docs/example_2_sol.json)
```
jcoupey, line 61: I would remove this step in order to not duplicate the check across workflows. Breaking the expected example output would still break the other workflow, but this one would only fail in case of a problem with `libosrm`.

.github/workflows/vroom.yml
```diff
@@ -0,0 +1,26 @@
...
+    branches:
+      - master
+  pull_request:
```
jcoupey, line 6: It would be great to avoid triggering a build when only markdown files are changed. Looks like `paths-ignore` would do that?

The typical situation is a single commit to just add a changelog entry.

jcoupey (commented):
This is great overall and will enhance the `libosrm`-related checks. Also I didn't know anything about GA but I find syntax easier to handle.

Could you add entries to `CHANGELOG.md` referencing the changes?

.github/workflows/libosrm.yml
Orthae: Added tags trigger.

.github/workflows/vroom.yml
Orthae: Same as above.

.github/workflows/libosrm.yml
Orthae: Removed path restriction on push to master.

.github/workflows/libosrm.yml
Orthae: Step removed.

.github/workflows/vroom.yml
Orthae: Added markdown files(.md) to ignore.

jcoupey:
@Orthae thanks again for submitting this, the first actions were triggered as expected upon merge and I removed the previous Travis config. :+1:

PR #545, opened 2021-07-13
Feature/421

### LINKED ISSUE #421: Set max number of tasks at vehicle level
We are setting 5 as the capacity but the problem is the system is assigning 5 delivery task and 5 pickup task separately which means 10 task at total. Is there any solution for set capacity regardless of task type(pickup and delivery)?

**input:**
```
{
  "vehicles": [
    {
      "id": 1,
      "start": [
        2.35044,
        48.71764
      ],
      "end": [
        2.35044,
        48.71764
      ],
      "capacity": [
        1
      ]
    },
    {
      "id": 2,
      "start": [
        2.35044,
        48.71764
      ],
      "end": [
        2.35044,
        48.71764
      ],
      "capacity": [
        1
      ]
    }
  ],
  "jobs": [
    {
      "id": 1,
      "service": 300,
      "delivery": [
        1
      ],
      "location": [
        1.98935,
        48.701
      ]
    },
    {
      "id": 2,
      "service": 300,
      "pickup": [
        1
      ],
      "location": [
        1.98935,
        48.701
      ]
    },
    {
      "id": 5,
      "service": 300,
      "delivery": [
        1
      ],
      "location": [
        1.98935,
        48.701
      ]
    },
    {
      "id": 6,
      "service": 300,
      "delivery": [
        1
      ],
      "location": [
        1.98935,
        48.701
      ]
    }
  ]
}
```

**output:**
```
{
        "success": true,
        "status": 200,
        "data": {
            "code": 0,
            "summary": {
                "cost": 0,
                "unassigned": 1,
                "delivery": [
                    2
                ],
                "amount": [
                    2
                ],
                "pickup": [
                    1
                ],
                "service": 900,
                "duration": 0,
                "waiting_time": 0,
                "priority": 0,
                "computing_times": {
                    "loading": 4,
                    "solving": 1
                }
            },
            "unassigned": [
                {
                    "id": 6,
                    "location": [
                        1.98935,
                        48.701
                    ]
                }
            ],
            "routes": [
                {
                    "vehicle": 1,
                    "cost": 0,
                    "delivery": [
                        1
                    ],
                    "amount": [
                        1
                    ],
                    "pickup": [
                        1
                    ],
                    "service": 600,
                    "duration": 0,
                    "waiting_time": 0,
                    "priority": 0,
                    "steps": [
                        {
                            "type": "start",
                            "location": [
                                2.35044,
                                48.71764
                            ],
                            "load": [
                                1
                            ],
                            "arrival": 0,
                            "duration": 0
                        },
                        {
                            "type": "job",
                            "location": [
                                1.98935,
                                48.701
                            ],
                            "id": 1,
                            "service": 300,
                            "waiting_time": 0,
                            "job": 1,
                            "load": [
                                0
                            ],
                            "arrival": 0,
                            "duration": 0
                        },
                        {
                            "type": "job",
                            "location": [
                                1.98935,
                                48.701
                            ],
                            "id": 2,
                            "service": 300,
                            "waiting_time": 0,
                            "job": 2,
                            "load": [
                                1
                            ],
                            "arrival": 300,
                            "duration": 0
                        },
                        {
                            "type": "end",
                            "location": [
                                2.35044,
                                48.71764
                            ],
                            "load": [
                                1
                            ],
                            "arrival": 600,
                            "duration": 0
                        }
                    ]
                },
                {
                    "vehicle": 2,
                    "cost": 0,
                    "delivery": [
                        1
                    ],
                    "amount": [
                        1
                    ],
                    "pickup": [
                        0
                    ],
                    "service": 300,
                    "duration": 0,
                    "waiting_time": 0,
                    "priority": 0,
                    "steps": [
                        {
                            "type": "start",
                            "location": [
                                2.35044,
                                48.71764
                            ],
                            "load": [
                                1
                            ],
                            "arrival": 0,
                            "duration": 0
                        },
                        {
                            "type": "job",
                            "location": [
                                1.98935,
                                48.701
                            ],
                            "id": 5,
                            "service": 300,
                            "waiting_time": 0,
                            "job": 5,
                            "load": [
                                0
                            ],
                            "arrival": 0,
                            "duration": 0
                        },
                        {
                            "type": "end",
                            "location": [
                                2.35044,
                                48.71764
                            ],
                            "load": [
                                0
                            ],
                            "arrival": 300,
                            "duration": 0
                        }
                    ]
                }
            ]
        }
    }
```

bottorezhan:
## Issue

Fixes #421 

## Tasks

- [x] Handle new input
- [x] Make sure the new constraint is used during the heuristic process
- [x] Make sure the new constraint is used during the local search process
- [x] Update docs/API.md
- [x] Update CHANGELOG.md
- [x] review

### DIFF
--- a/src/algorithms/heuristics/heuristics.cpp
+++ b/src/algorithms/heuristics/heuristics.cpp
@@ -132,6 +132,10 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
           continue;
         }
 
+        if (!current_r.is_valid_addition_for_max_tasks(input, job_rank)) {
+          continue;
+        }
+
         bool is_pickup = (input.jobs[job_rank].type == JOB_TYPE::PICKUP);
 
         bool try_validity = false;
@@ -237,6 +241,10 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
         }
 
         if (input.jobs[job_rank].type == JOB_TYPE::DELIVERY) {
+          continue;
+        }
+
+        if (!current_r.is_valid_addition_for_max_tasks(input, job_rank)) {
           continue;
         }
 
@@ -504,6 +512,10 @@ T dynamic_vehicle_choice(const Input& input, INIT init, float lambda) {
           continue;
         }
 
+        if (!current_r.is_valid_addition_for_max_tasks(input, job_rank)) {
+          continue;
+        }
+
         bool is_pickup = (input.jobs[job_rank].type == JOB_TYPE::PICKUP);
 
         bool try_validity = false;
@@ -610,6 +622,10 @@ T dynamic_vehicle_choice(const Input& input, INIT init, float lambda) {
         }
 
         if (input.jobs[job_rank].type == JOB_TYPE::DELIVERY) {
+          continue;
+        }
+
+        if (!current_r.is_valid_addition_for_max_tasks(input, job_rank)) {
           continue;
         }
 
--- a/src/algorithms/heuristics/heuristics.cpp
+++ b/src/algorithms/heuristics/heuristics.cpp
@@ -76,16 +76,18 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
   // within the heuristic.
   std::vector<Index> vehicles_ranks(nb_vehicles);
   std::iota(vehicles_ranks.begin(), vehicles_ranks.end(), 0);
-  // Sort vehicles by "higher" capacity or by time window in case of
-  // capacities ties.
+  // Sort vehicles by decreasing max number of tasks allowed, then
+  // capacity (not a total order), then working hours length.
   std::stable_sort(vehicles_ranks.begin(),
                    vehicles_ranks.end(),
                    [&](const auto lhs, const auto rhs) {
                      auto& v_lhs = input.vehicles[lhs];
                      auto& v_rhs = input.vehicles[rhs];
-                     return v_rhs.capacity << v_lhs.capacity or
-                            (v_lhs.capacity == v_rhs.capacity and
-                             v_lhs.tw.length > v_rhs.tw.length);
+                     return v_lhs.max_tasks > v_rhs.max_tasks or
+                            (v_lhs.max_tasks == v_rhs.max_tasks and
+                             (v_rhs.capacity << v_lhs.capacity or
+                              (v_lhs.capacity == v_rhs.capacity and
+                               v_lhs.tw.length > v_rhs.tw.length)));
                    });
 
   auto costs = get_jobs_vehicles_costs(input);
--- a/src/algorithms/heuristics/heuristics.cpp
+++ b/src/algorithms/heuristics/heuristics.cpp
@@ -134,7 +134,8 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
 
         bool is_pickup = (input.jobs[job_rank].type == JOB_TYPE::PICKUP);
 
-        if (current_r.size() + (is_pickup ? 2 : 1) > vehicle.max_number_of_tasks) {
+        if (current_r.size() + (is_pickup ? 2 : 1) >
+            vehicle.max_number_of_tasks) {
           continue;
         }
 
@@ -243,8 +244,9 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
         if (input.jobs[job_rank].type == JOB_TYPE::DELIVERY) {
           continue;
         }
-   
-        if (input.jobs[job_rank].type == JOB_TYPE::SINGLE and current_r.size() + 1 <= vehicle.max_number_of_tasks) {
+
+        if (input.jobs[job_rank].type == JOB_TYPE::SINGLE and
+            current_r.size() + 1 <= vehicle.max_number_of_tasks) {
           for (Index r = 0; r <= current_r.size(); ++r) {
             float current_add = utils::addition_cost(input,
                                                      job_rank,
@@ -269,7 +271,8 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
           }
         }
 
-        if (input.jobs[job_rank].type == JOB_TYPE::PICKUP and current_r.size() + 2 <= vehicle.max_number_of_tasks) {
+        if (input.jobs[job_rank].type == JOB_TYPE::PICKUP and
+            current_r.size() + 2 <= vehicle.max_number_of_tasks) {
           // Pre-compute cost of addition for matching delivery.
           std::vector<Gain> d_adds(current_r.route.size() + 1);
           std::vector<unsigned char> valid_delivery_insertions(
@@ -510,7 +513,8 @@ T dynamic_vehicle_choice(const Input& input, INIT init, float lambda) {
 
         bool is_pickup = (input.jobs[job_rank].type == JOB_TYPE::PICKUP);
 
-        if (current_r.size() + (is_pickup ? 2 : 1) > vehicle.max_number_of_tasks) {
+        if (current_r.size() + (is_pickup ? 2 : 1) >
+            vehicle.max_number_of_tasks) {
           continue;
         }
 
@@ -621,7 +625,8 @@ T dynamic_vehicle_choice(const Input& input, INIT init, float lambda) {
           continue;
         }
 
-        if (input.jobs[job_rank].type == JOB_TYPE::SINGLE and current_r.size() + 1 <= vehicle.max_number_of_tasks) {
+        if (input.jobs[job_rank].type == JOB_TYPE::SINGLE and
+            current_r.size() + 1 <= vehicle.max_number_of_tasks) {
           for (Index r = 0; r <= current_r.size(); ++r) {
             float current_add = utils::addition_cost(input,
                                                      job_rank,
@@ -646,7 +651,8 @@ T dynamic_vehicle_choice(const Input& input, INIT init, float lambda) {
           }
         }
 
-        if (input.jobs[job_rank].type == JOB_TYPE::PICKUP and current_r.size() + 2 <= vehicle.max_number_of_tasks) {
+        if (input.jobs[job_rank].type == JOB_TYPE::PICKUP and
+            current_r.size() + 2 <= vehicle.max_number_of_tasks) {
           // Pre-compute cost of addition for matching delivery.
           std::vector<Gain> d_adds(current_r.route.size() + 1);
           std::vector<unsigned char> valid_delivery_insertions(
--- a/src/algorithms/heuristics/heuristics.cpp
+++ b/src/algorithms/heuristics/heuristics.cpp
@@ -134,8 +134,9 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
 
         bool is_pickup = (input.jobs[job_rank].type == JOB_TYPE::PICKUP);
 
-        if (current_r.size() + (is_pickup ? 2 : 1) > vehicle.max_number_of_tasks)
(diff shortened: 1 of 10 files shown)


jcoupey:
Thanks for submitting a PR! I suggest we move the technical questions to this thread. Concerning https://github.com/VROOM-Project/vroom/issues/421#issuecomment-878085542:

- you're correct in assuming that all `Intra*` operators won't change route size;
- there is no real explanation/documentation for the other operators beside the ones described in the slide, I'll provide one here as soon as I can.

jcoupey:
So here is a quick description of local search operators.

## The ones we don't care about

- All `Intra*` operators (`IntraCrossExchange`, `IntraExchange`, `IntraMixedExchange`, `IntraOrOpt`, `IntraRelocate`) are variants that only apply to a single route. As already pointed out, they don't modify the route size so are out of scope here.
- `UnassignedExchange`: for all unassigned jobs and all routes, it tries to swap the unassigned job with a job already assigned in the current route. We only do so if it's better in term of priority score, or cost if priority does not change. When applied, this move does not change the route size.

## The documented ones

`Relocate`, `Exchange`, `OrOpt`, `CrossExchange`, `TwoOpt` and `ReverseTwoOpt` have a self-explanatory picture description at https://github.com/VROOM-Project/vroom/issues/421#issuecomment-878085542, feel free to ask for more details if required.

## The rest

The remaining operators do impact route size for vehicles so they are relevant here.

- `RouteExchange` simply swaps routes between source and target vehicles. It's only applied when this can make a difference, that is when vehicles may have different start/end points or different profiles.
- `MixedExchange` is a mix between `Exchange` and `CrossExchange` where one task from source route is swapped with two consecutive tasks from target route.
-  `PdShift` can be seen as a `Relocate` version for `shipment` objects: it moves two tasks (a pickup and the associated delivery, not necessarily next one to another) from source route to target route. This operator computes the best valid insertion position for both tasks in target route along the way.

## Note on operators

We try to be clever on what checks we apply with operators. For example, `is_valid` being usually more expensive, it's only called when potential gain is known to be higher than the best one we already have at hand. Sometimes we don't even call `gain` when a faster `gain_upper_bound` pre-computation does discard the operator. In many other situations, we have checks to abort early without even creating the operator object, e.g. for `skills` incompatibility. This is especially relevant as we have lots of nested `for` loops.

The reason why I'm writing this is to explain that there is quite some boilerplate around the use of operators in `local_search.cpp`, and that's where the filtering should happen whenever we spot that an operator would lead to a break in max number of tasks.

src/algorithms/heuristics/heuristics.cpp
```diff
@@ -132,6 +132,10 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
           continue;
         }
 
+        if (!current_r.is_valid_addition_for_max_tasks(input, job_rank)) {
```
jcoupey, line 135: I'd be in favor of performing an inline explicit check here based on vehicle `max_number_of_tasks` and size of `current_r` and remove the `RawRoute::is_valid_addition_for_max_tasks` member altogether.

Having the test logic factored in a single function sounds appealing in the first place but:
- we still need to use explicit conditions elsewhere (cf the filters already applied for local search operators);
- an explicit test like `v.max_number_of_tasks == current_r.size()` (for single jobs) feels as expressive as a function call.

bottorezhan: In most cases where `RawRoute::is_valid_addition_for_max_tasks` method is used, type of job is not filted (can be both `JOB_TYPE::SINGLE` and `JOB_TYPE::PICKUP`)

jcoupey: Yes, I noticed that. The test could then rely on the `is_pickup` value which is defined just below. In some other places for this file there is an `if` branching on the job type so the appropriate test for each case could be added there.

bottorezhan: Was busy with work, removed `RawRoute::is_valid_addition_for_max_tasks` and added `max_number_of_tasks` to docs/API.md

src/utils/input_parser.cpp
```diff
@@ -122,6 +122,21 @@ inline Duration get_priority(const rapidjson::Value& object) {
...
+  size_t max_number_of_tasks = std::numeric_limits<size_t>::max();
+  if (object.HasMember("max_number_of_tasks")) {
+    if (!object["max_number_of_tasks"].IsUint()) {
+      throw Exception(ERROR::INPUT, "Invalid max_number_of_tasks value.");
+    }
+    max_number_of_tasks = object["max_number_of_tasks"].GetUint();
+    if (!max_number_of_tasks) {
+      throw Exception(ERROR::INPUT, "Invalid max_number_of_tasks value.");
+    }
+
```
jcoupey, line 135: I'm curious about when this would be triggered? For example in the similar `get_service`, we don't have that check at all.

src/algorithms/heuristics/heuristics.cpp
```diff
@@ -134,6 +134,9 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
 
         bool is_pickup = (input.jobs[job_rank].type == JOB_TYPE::PICKUP);
 
+        if (current_r.size() + (is_pickup ? 2 : 1) > vehicle.max_number_of_tasks)
```
jcoupey, line 137: Nitpicking: please use braces around the `if` clause.

src/algorithms/heuristics/heuristics.cpp
```diff
@@ -506,6 +510,9 @@ T dynamic_vehicle_choice(const Input& input, INIT init, float lambda) {
...
```
jcoupey, line 513: Nitpicking: please use braces around the `if` clause.

src/algorithms/local_search/local_search.cpp
```diff
@@ -343,6 +343,14 @@ void LocalSearch<Route,
         if (addition_cost == std::numeric_limits<Gain>::max()) {
           continue;
         }
+
+        const auto& current_r = _sol[routes[i]];
+        const auto& vehicle = _input.vehicles[current_r.vehicle_rank];
+        bool is_pickup = (_input.jobs[j].type == JOB_TYPE::PICKUP);
+
+        if (current_r.size() + (is_pickup ? 2 : 1) > vehicle.max_number_of_tasks)
```
jcoupey, line 351: Nitpicking: please use braces around the `if` clause.

jcoupey (changes requested):
Thanks for the follow-up work, we now have all operators covered! I made a few notes on some checks.

Bonus point if you can also run the formatting script. ;-)

src/utils/input_parser.cpp
bottorezhan: In my honest opinion, `max_number_of_tasks` should be a positive number or not provided

src/algorithms/local_search/local_search.cpp
```diff
@@ -758,6 +771,16 @@ void LocalSearch<Route,
...
           }
 
+          const auto& s_v = _input.vehicles[_sol[s_t.first].vehicle_rank];
+          const auto& t_v = _input.vehicles[_sol[s_t.second].vehicle_rank];
+
+          if (s_rank - t_rank + _sol[s_t.second].size() >
+                s_v.max_number_of_tasks or
+              t_rank - s_rank + _sol[s_t.first].size() >
+                t_v.max_number_of_tasks) {
+            continue;
```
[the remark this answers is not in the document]
jcoupey: Not a good idea on second though because it would overflow if `max_number_of_tasks` is `std::numeric_limits<size_t>::max()`. Maybe putting braces around `(_sol[s_t.second].size() - t_rank)`, guaranteed to be non-negative?

src/utils/input_parser.cpp
jcoupey: Agreed, my point is that we already check that by throwing if `max_number_of_tasks` is provided and `object["max_number_of_tasks"].IsUint()` is false.

src/algorithms/heuristics/heuristics.cpp
```diff
@@ -240,7 +243,8 @@ template <class T> T basic(const Input& input, INIT init, float lambda) {
           continue;
         }
 
-        if (input.jobs[job_rank].type == JOB_TYPE::SINGLE) {
+
+        if (input.jobs[job_rank].type == JOB_TYPE::SINGLE and current_r.size() + 1 > vehicle.max_number_of_tasks) {
```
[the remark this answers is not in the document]
bottorezhan: oops, fixed !

src/utils/input_parser.cpp
bottorezhan: Ok, didn't know. I will remove it !

bottorezhan: Removed !

src/algorithms/local_search/local_search.cpp
```diff
@@ -665,6 +674,11 @@ void LocalSearch<Route,
           continue;
         }
 
+        const auto& v_t = _input.vehicles[s_t.second];
+        if (_sol[s_t.first].size() + 1 > v_t.max_number_of_tasks) {
```
jcoupey, line 678: The right-hand side change is missing: we should compare to `v_s.max_number_of_tasks` where `v_s` is the source vehicle (const ref defined using `s_t.first`).

src/algorithms/local_search/local_search.cpp
```diff
@@ -758,6 +772,16 @@ void LocalSearch<Route,
             continue;
           }
 
+          const auto& s_v = _input.vehicles[s_t.first];
+          const auto& t_v = _input.vehicles[s_t.second];
+
+          if (s_rank + _sol[s_t.second].size() >
+                t_rank + s_v.max_number_of_tasks or
+              t_rank + _sol[s_t.first].size() >
+                s_rank + t_v.max_number_of_tasks) {
```
jcoupey, line 781: My bad. This was actually not a good idea since `t_rank + s_v.max_number_of_tasks` will overflow whenever number of tasks is `std::numeric_limits<size_t>::max()`. Maybe we could keep the previous version but reorder and group computation for values that we know are positive: `(_sol[s_t.second].size() - t_rank)` and `(_sol[s_t.first].size() - s_rank)`.

src/algorithms/local_search/local_search.cpp
```diff
@@ -1330,6 +1387,11 @@ void LocalSearch<Route,
                           return sum + _sol_state.route_costs[c];
                         });
       assert(new_cost + best_gain == previous_cost);
+      
+      for (auto v_rank : update_candidates) {
+        assert(_sol[v_rank].size() <= _input.vehicles[v_rank].max_number_of_tasks);
```
jcoupey, line 1392: Useful addition! We could avoid this loop altogether by moving the assert to the similar loop right above (line 1379).

jcoupey (changes requested):
Thanks for the fixes. Just a couple more comments.

jcoupey:
Looks good to me, I'll have a go at testing this with actual instances as soon as I can.

If everything is fine, we only need a changelog entry and running [the formatting script](https://github.com/VROOM-Project/vroom/blob/master/CONTRIBUTING.md#automatic-formatting). If you don't want to handle that, I can do it, just let me know.

bottorezhan:
It will be better if you will handle it!

Thank you!

jcoupey:
OK, then I'll test and do the final adjustments before merging. Thanks for your work on this!

bottorezhan:
Hi, any progress ?

jcoupey (approved):

jcoupey:
Thanks for your work @bottorezhan!

Issue #358, opened 2020-06-26
Warehouse access time


boooch:
In case of  using pickup and delivery, part or all pickups can be in one point (warehouse). If I set service time fop pickup for example 10 minutes in case of 5 shipments - full service time will be 50 minutes. But this is not real situation. Best solution IMHO is to add additional variable to job - access time. In this case I can set access time for warehouse - 10 minutes and service time for each shipment - 1 minute, so we will get 11 minutes in case of 1 shipment and 15 minutes in case of 5 shipments. What are you thinking about it? This is real use case

jcoupey:
This sounds really similar to #353 where a couple options/hacks were discussed.

jcoupey:
It's probably possible to achieve the expected behaviour without a huge change of the current implementation. The idea would be to keep the `service` value as a small unit of time that is actually required with an extra step, but add a `setup` time value that would account for the cost that should not be duplicated (e.g. arriving, parking etc.).

For a task at a new location, we would apply `setup + service` time, but then for any additional task at the same location, only the `service` time would be applied.

For example, 5 pickups done in a row with each a setup of 10 minutes and service of 2 minutes, then the overall time would be 10 + 5 * 2 = 20 minutes.

Provided it's always possible to access previous location for all checks/route changes without changing the current implemenation (I think so), then it would be a matter of spotting all places where a service time is applied to decide whether the setup time should be added.

leobudima:
Just to add anecdotally that this does prove to be an inconvenience in different situations - e.g.:

-  regular delivery of multiple tasks at the same location, when because of accumulated service time at location other possibly doable tasks are unassigned

- modeling tasks as shipments for the purpose of reloading at depot, and then having a huge accumulated service time at depot if not specifically managing pickup service duration differently

and the `setup + service` sounds like a great feature!

I'm sure people will end up asking about different setup times for different types of tasks etc, but realistically, a simple, "optimization-wide" `setup` value is more than great for majority of cases.

jcoupey:
@leobudima thanks for your input. I agree this is significantly important and we need to have it at some point.

I know I won't have time to commit to this feature any time soon, so for the time being I'd welcome any contribution. This is why I described a possible implementation above and can provide more pointers for those interested. I'd gladly merge a PR that implements this idea without increasing solving complexity.

leobudima:
@jcoupey thanks for your reply! 

[...]

PR #547, opened 2021-07-18
Add setup time

jcoupey:
## Issue

This PR aims at fixing #358 by adding a new `setup` duration for tasks that would only be applied when tasks occur at a new location. In particular for any task happening at the same location as the previous one, only `service` time would apply.

## Tasks

 - [x] Parse the `setup` value in input
 - [x] Adjust the `Job` object and ctor to store it
 - [x]  Apply `setup` time to new location tasks
 - [x] Adjust validity check logic
 - [x] Store `setup` value in relevant objects used to describe a solution
 - [x] Report `setup` time in `route.steps` objects
 - [x] Report cumulated setup values at route and summary level in output
 - [x] Apply `setup` time when necessary in `plan` mode
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
I've mostly been doing some refactoring of the time window handling logic so far. The good news is that defining new data structures to hold various information and pass it around in a better way seems to significantly reduce the overhead. On the usual Solomon bench with TW, I've noticed a speedup of around 20 to 30% on the tests I've made between the tip of this PR and the parent commit on `master`. The underlying logic for validity checks is (hopefully) unchanged so far.

Now on `setup` time: deciding a some point (say when adding a job in a route) whether to apply `setup` time or not depends on the previous task location. I've included that in `PreviousInfo.location_index` so we now have it handy. The problem is that deciding the earliest date for the new task requires to use the earliest date for the previous task (OK, stored) and the overall "action" time for previous task. Currently the "action" time is only service, but now it will actually depend on whether we apply setup time to the previous task, which in turn depends on etc.
I think we can get away with this to still decide whether to apply `setup` time in constant time (algorithmically speaking) simply by storing combined "action" time (`service + setup`) for all tasks in routes, just like we store earliest and latest dates.

jcoupey:
I've come across a tricky limitation of the current TW validity check logic in `TWRoute::is_valid_addition_for_tw` (the one that replaces a range of jobs). In some situations, it returns `true` for a change but yet the subsequent call to `TWRoute::replace` applying the change fails after updating the range and upon updating subsequent earliest times with `TWRoute::fwd_update_earliest_from`.

After investigating, I think I now what's going on. Say the first task `t` after the replace range used to be at the same location as  previous (removed) task(s). Then the setup duration that was applied before to one of the tasks before `t` now happens **after** arrival at `t`. The check performed for validity is that the new earliest date for `t` is lower than the current latest date for `t`. This may be `true` but at the same time appending a new setup time after arrival at `t`shifts earliest dates for subsequent jobs that may not be doable any more.

jcoupey:
Another limitation: say we have a route `*[range]A*` where `[range]` is replaced by a set of other tasks (or none) in a call to `TWRoute::replace`. With our current logic, task `A` is the first one in the route whose previous latest date is still valid. So what we currently do is call `TWRoute::bwd_update_latest_from` starting at `A` in order to update all previous latest dates backward. This works when the time spent for a given task never changes.

Now depending on `A`s previous location before and after replacing `[range]`, the action time at `A` may change, either by applying new setup time or removing it. This means that the previous latest date for `A` is not valid any more. What we should do in that situation is call `bwd_update_latest_from` starting at the task after A (if any), or directly update the latest time at `A` before backward propagation.

jcoupey:
My remaining troubles with the TW logic should now be automagically solved by including the changes from #558 via a `master` merge.

PR #555, opened 2021-08-20
Custom cost matrices

### LINKED ISSUE #415: Supply cost matrix and duration matrix separately
Hi,

We provide VROOM a customized cost matrix rather than letting VROOM fetch from OSRM (As it is already supported by VROOM). But manipulating the cost matrix deforms time windows due to durations are the same as the cost matrix. It'd be nice if we could pass a duration matrix along with a cost matrix. Then VROOM could optimize the cost matrix but check time windows by looking up to the duration matrix.

Thanks!

jcoupey:
## Issue

This PR aims at pushing the separation between travel times and costs one step further by allowing to pass custom cost matrices in input. Fixes #415.

## Tasks

 - [x] Parse `matrices.*.costs` matrices in input
 - [x] Adjust `vroom::Input` to hold the various types of matrices
 - [x] Update `vroom::CostWrapper` to introduce various ways to compute/use costs depending on context
 - [x] Define proper cost values when generating `vroom::Route` objects in solutions
 - [x] Fix durations/costs separation leftovers
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
Fixes #570

jcoupey:
While testing, I'm seeing situations where this PR currently hits:

https://github.com/VROOM-Project/vroom/blob/f7dfa76ded51f96f45dc6a8f9467efe684cf0f40/src/algorithms/local_search/local_search.cpp#L1388

This happens when custom costs and durations matrix are different. It means we have a faulty operator gain evaluation. I suspect this is a leftover from the costs/durations separation done a while back.

jcoupey:
We're getting there but I'm still seeing weird things going on. For example when using a cost matrix that is the duration matrix doubled, then for *some* instances the solving path is changed at some point and we get a different solution.

AFAICT this is related to a problem in `UnassignedExchange` gain evaluation. There are probably still a couple values involving durations instead of costs.

jcoupey:
After much digging, I think the last problems I'm seeing are not at all cost-related. My feeling is that comparing solutions across instances where we scale the costs is only a way to highlight a problem that is already present in the codebase, so I'll open a dedicated ticket for that.

PR #580, opened 2021-09-21
Implement SWAP*

### LINKED ISSUE #507: Add SWAP* operator
Our local search relies on various SWAP-like neighborhoods that will perform in-place exchanges of tasks (or pair of tasks), either between two different routes, or in the same route. For example the Exchange operator typically will spot situations like the following (edges are gray before applying and red after):

exchange_after

But requiring that both jobs take each other's place in the other route is very limiting and we're missing situations where the real gain would happen when moving tasks across routes *while* changing their respective ranks. Probably this can happen sometimes during the current search with two subsequent relocate moves but again in a very limited manner: the intermediate state has to be both valid (capacity, TW) and interesting in term of gain.

The drawback of this kind of move is that the *a priori* complexity is `O(n^4)` where `n` is the typical route length. Fortunately, [this recent paper](https://arxiv.org/abs/2012.10384) from Thibaut Vidal shares a clever way to reduce the required exploration, yet making sure to spot the best new insertion ranks. This brings down the operator lookup complexity to `O(n^2)`.

So it appears we could have our cake and eat it too by superseding the existing Exchange operator with SWAP*: we'd then be able to reach out to more relevant neighboring solutions while not increasing the overall complexity.

jcoupey:
## Issue

Fixes #507 

## Tasks

 - [x] Implement spotting best SWAP* option
 - [x] First working implementation for `cvrp::SwapStar` operator
 - [x] First working implementation for `vrptw::SwapStar` operator
 - [x] Adjust paper approach to account for capacity constraints
 - [x] Adjust paper approach to account for TW constraints
 - [x] Expand search to less interesting available moves if best SWAP* option is not valid
 - [x] Benchmark
 - [x] Deprecate `Exchange` code
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
# Note on benchmarks

 TL;DR superseding the usual Exchange operator with the new SWAP*-adjusted implementation does improve results in most cases. This means generally reducing average/median/worst gap to optimal while stretching computing time by around 5 to 10% usually, which is definitely acceptable for the sake of quality improvement.

Results vary across exploration levels and benchmarks classes so I'll just report general trends.

## CVRP

- More instances solved to full number of jobs
- Steady reduction of average optimal gap around -0.5 / -0.6 point

This may not sound like a big deal but it is significant since gaps are already quite low. For example with `-x 5`, average gap drops from 1.79% to 1.32% while median gap drops from 1.05% down to 0.68%.

## MDVRP and HVRP

Not that many instances so probably not very significant, but same trend as above with more volatility in average gap reduction (from ~-0.1 to -0.8 point).

## VRPTW

Steady reduction in costs for Homberger instances while average gap actually slightly increases for Solomon instances. The latter is a bit frustrating and a closer inspection shows that most Solomon instances are actually slightly improved but a few of them (in the R and RC classes) generate quite poorer solutions. I did not come up with a full explanation for this, other than the fact that the "random" instances are tougher to solve (especially with tight constraints) so any local search change is more prone to lead to getting trapped in a worse local minima.

src/algorithms/local_search/swap_star_utils.h
```diff
@@ -0,0 +1,353 @@
...
+
+      assert(swap_choice_options.size() <= 16);
+
+      for (const auto& sc : swap_choice_options) {
+        // Browse interesting options by decreasing gain and check for
+        // validity.
+        const auto s_insert = get_insert_range(source.route,
+                                               s_rank,
+                                               target.route[t_rank],
+                                               sc.insertion_in_source);
```
jcoupey, line 249: I think this is the actual pain point for computing times since here `get_insert_range` populates a new `std::vector` that serve as insertion range for the validity checks down the line.

The reason this is required in the first place is that the actual range is mostly a part of the source route, but with the target node added either in front or back depending on respective insertion ranks.

I don't see a way to solve this better for now, but I hope there is. Maybe at some point using ranges to be able to iterate without actually building the range?

src/algorithms/local_search/swap_star_utils.h
```diff
@@ -0,0 +1,353 @@
...
+  // Search phase.
+  auto best_choice = empty_choice;
+  Gain best_gain = best_known_gain;
+
+  for (const auto& s_element : top_insertions_in_target) {
+    const auto s_rank = s_element.first;
+    const auto& target_insertions = s_element.second;
+
+    const auto source_remove_gain =
+      sol_state.node_gains[source.vehicle_rank][s_rank];
```
jcoupey, line 159: There is a flaw here because what the algorithm requires is a cost delta. Namely if job `j` has previous and next steps `p` and `n`, then `Delta(j, p, n) = cost(p, j) + cost(j, n) - cost(p, n)`.

This is exactly what `sol_state.node_gains` holds in most situations. The problem is if the route has length 1 and vehicle has start and end that differ. Then for the delta computation `cost(p, n)` should be the cost of getting from start to end in an empty route. This is not accounted for in `sol_state.node_gains` because we consider that an empty route does not generate a start -> end cost.

Currently this leads to wrong gain estimates with a SWAP* move involving such a route, which hits the assert on gain verification we have down the line after applying the moves.

PR #604, opened 2021-10-28
switch arg parser to cxxopts

### LINKED ISSUE #602: replace arg parser with cxxopts
I recently migrated valhalla's command line arg parser to https://github.com/jarro2783/cxxopts . while that was really necessary (using linked boost.program_options before), I'd still like to migrate vroom there if you're interested. it's much more legible and would replace the win get_opt.h stuff I introduced a while back.

nilsnolde:
fixes #602 

I poc'd it real quick, it's definitely not tested yet, but you'll see the way it'd be implemented.

a few questions from my side:
- the `heuristic` arg is now showing on `help`, I guess that's mostly for you to debug? I can't see a way to hide it in cxxopts, but also don't think it's that bad it's showing now
- maybe we can get rid of the `CLArgs` struct or even `cl_args.h` and pull the stuff into `main.cpp`? defaults are applied via cxxopts which is IMHO more idiomatic/expected. main is only passing single parameters to functions as far as I could see, so it wouldn't be a big change I think. anyways, that's maybe micro-"optimizing"..
- no git submodule? cxxopts repo has 500kB size, but rapidjson is a little bigger with 30 MB

## Tasks

 - [x] review


jcoupey:
Thanks for drafting this PR, I'd like to move on with this as part of the release after v1.11. On the questions you raised:

>  the heuristic arg is now showing on help, I guess that's mostly for you to debug?

Yep, that is really a debugging entry, that's why it is not actually documented as such. I guess we can always raise the question of optionally showing on the cxxopts repo, but no harm done if not possible.

> maybe we can get rid of the CLArgs struct or even cl_args.h and pull the stuff into main.cpp

That's an option, not sure what is the most readable/explicit between having everything straight into the main, or on the other hand having a dedicated placeholder for everything related to user input. In any case, we can always change that as another step, so let's stick to a simple replacement of the current behavior in this PR.

> no git submodule? cxxopts repo has 500kB size, but rapidjson is a little bigger with 30 MB

I agree versioning dependencies along with the core code for this repo is not ideal so the whole idea of submodules is appealing. On the other hand the current `include/rapidjson` folder is 652K so pulling 30MB sounds weird.
What if we introduce cxxopts as a submodule and evaluate the switch to submodules separately for other dependencies (polylineencoder and rapidjson)?

nilsnolde:
jep, agree on all points. will do a git submodule and leave the rest for now. skimming over some issues in cxxopts, I found out how to hide certain options, will implement that for the heuristic arg.

nilsnolde:
- added `cxxopts` as submodule
- hide debug args from output on `vroom -h`

for me this is done. no stress with a review. also needs some more testing, I just did example_2.json. would be good if you could eventually try this branch with your common arg combos @jcoupey

jcoupey:
Thanks @nilsnolde. I can't really commit to reviewing soon but I definitely want to include this early in the next iteration so we have plenty of time to test from `master` and feature branches.

nilsnolde:
yeah no worries, we're not depending on this, it has infinite time from our side:)

src/main.cpp
```diff
@@ -12,121 +12,83 @@ All rights reserved (see LICENSE).
...
-      nb_threads_arg = optarg;
-      break;
-    case 'x':
-      exploration_level_arg = optarg;
-      break;
-    default:
-      break;
-    }
-    opt = getopt(argc, argv, optString);
+  // clang-format off
```
jcoupey, line 38: I agree clang-format is especially ugly on the `options....` part, but what about:

- still clang-formatting the `options` definition and the `add_options("debug_group")` part;
- avoid overly long lines by using one argument per line for some of the options.

src/main.cpp
```diff
@@ -12,121 +12,74 @@ All rights reserved (see LICENSE).
...
-    std::cerr << "[Error] " << message << std::endl;
-    vroom::io::write_to_json({error_code, message}, false, cl_args.output_file);
-    exit(error_code);
+  // parse and update some params
+  vroom::io::update_host(cl_args.servers, host_arg);
+  vroom::io::update_port(cl_args.servers, port_arg);
+  if (!limit_arg.empty()) {
+    // Internally timeout is in milliseconds.
+    cl_args.timeout = 1000 * std::stof(limit_arg);
   }
```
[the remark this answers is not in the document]
jcoupey: Was just about to mention that for `-x`, `-l` and `-t`. ;-)

Note that the exception handling somehow changed meanwhile in #641. You may want to rebase or merge current `master` before adding the `try` back.

src/structures/cl_args.cpp
```diff
@@ -17,13 +17,7 @@ namespace io {
...
 
 // Default values.
-CLArgs::CLArgs()
-  : check(false),
-    geometry(false),
-    router(ROUTER::OSRM),
-    nb_threads(4),
-    exploration_level(5) {
-}
+CLArgs::CLArgs(){}
```
jcoupey, line 20: We can probably remove this altogether from both the cpp and header file and let the compiler generate a default ctor.

jcoupey (commented):
I didn't spend much time testing so far but went through the changes and left a few comments.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "include/cxxopts"]
+	path = include/cxxopts
+	url = https://github.com/jarro2783/cxxopts
```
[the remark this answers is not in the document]
nilsnolde: jep, that's the tagged commit for 3.0.0 ;) just happened to be the latest at the time of submoduling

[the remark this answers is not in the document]
nilsnolde: https://github.com/jarro2783/cxxopts/tags

src/main.cpp
nilsnolde: hm. from my experience that's not how clang-format works, i.e. you can't just set up your own formatting here (one arg per line), would also kinda defeat the whole purpose of it I guess. I couldn't verify since I'm on arch linux and have only v13, not v10 available. the last commit has the changes you proposed, could you run and try for yourself? 

side note, in valhalla we have a kinda decent way to detach the clang-format version from individual package managers: https://github.com/jarro2783/cxxopts/tags

nilsnolde:
@jcoupey addressed your comments for the most part. probably we don't get around turning off clang-format, but happy to see if you can verify/dismiss my thoughts locally.

src/main.cpp
```diff
@@ -12,14 +12,13 @@ All rights reserved (see LICENSE).
 #include <sstream>
 #ifndef _WIN32
 #include <unistd.h>
```
jcoupey, line 14: This is not needed any more either?

src/main.cpp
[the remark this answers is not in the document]
jcoupey: Passing garbage to `-l` still breaks.

src/main.cpp
```diff
@@ -12,121 +12,83 @@ All rights reserved (see LICENSE).
...
-#else
-#include "getopt_win.h"
 #endif
 
 #if USE_LIBOSRM
 #include "osrm/exception.hpp"
 #endif
 
+#include "../include/cxxopts/include/cxxopts.hpp"
 #include "problems/vrp.h"
```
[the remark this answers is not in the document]
jcoupey: Up ;-)

jcoupey (changes requested):
The overall changes look OK from what I can tell but I get puzzling results when running `$ vroom -i docs/example_1.json`

I sometimes get:
```
[Error] VROOM compiled without libosrm installed.
{"code":3,"error":"VROOM compiled without libosrm installed."}
```
while the error used to be about not being able to reach `osrm-routed` (maybe a change in default router).

Also I sometimes randomly get a segfault for that same command instead of the above error.

krypt-n:
> Also I sometimes randomly get a segfault for that same command instead of the above error.

It appears that `cl_args.router` is never set if `router_arg == "osrm"` (the default) and thus contains uninitialized memory

src/main.cpp
nilsnolde: jep true:)

nilsnolde:
ok resolved the outstanding issues I hope. thanks @krypt-n , you're right of course.

github seems to have a bug.. I can't comment on half of your review comments @jcoupey , so:
- `-l` was a string before, I changed to `unsigned`, now garbage is properly handled
- improved the error reporting: cxxopts tells the user the argument which failed but no context, so I added the previous (before this PR) error message
- `cl_args.router` gets a default `osrm` value
- I didn't quite understand the `Up ;-)` comment though.. above all other includes is what you'd like to see or?

I checked a few commands and it worked fine for me now, in terms of examples only `example_2.json` though. let me know if you find more problems.

nilsnolde:
oh and still no formatting..

src/main.cpp
```diff
@@ -57,8 +54,8 @@ int main(int argc, char** argv) {
       cxxopts::value<bool>(cl_args.geometry)->default_value("false"))
     ("i,input-file", "Read input from 'input-file' rather than from stdin",
       cxxopts::value<std::string>(cl_args.input_file))
-    ("l,limit", "Stop solving process after 'limit' seconds",
-      cxxopts::value<std::string>(limit_arg)->default_value(""))
+    ("l,limit", "Stop solving process after 'limit' seconds. Default 0, i.e. unlimited.",
+      cxxopts::value<unsigned>(limit_arg)->default_value("0"))
```
jcoupey, line 58: This won't work since 0 is a valid value for `-l` which does not mean "no limit" but result in 0 time left for local search. You can probably keep the previous logic of using a string (default empty) and calling `std::stof` with a try block if the string is not empty.

jcoupey (changes requested):
Thanks for the changes, I think the only remaining concern is about the `-l` option and default value.

src/main.cpp
nilsnolde: ah pity:) if `-1` (and checking < 0) is not an option for you I'll use the string strategy again

nilsnolde: argh unsigned haha I'll go for string

nilsnolde: can't say it got prettier with nested try/catch but found it still to be more elegant solution. ok for you?

nilsnolde:
@jcoupey should we have another round so we can merge this?

actually this is (kinda) blocking windows support in the python bindings as well (https://github.com/VROOM-Project/pyvroom/issues/45). let me know if there's anything left to do and we can get it out of our heads:)

nilsnolde:
turns out that's a bit of a tricky case.. since cxxopts reserves commas to split arguments, e.g. `--location berlin,paris` will lead to `args.location = {"berlin", "paris"}` but at the same time supports having the same option multiple times, so the result is that `--location berlin,paris --location london` will be parsed to a flat vector `{"berlin,paris,london"}`, not `{{"berlin","paris},{"london"}}`.

I had to change cxxopts `CXXOPTS_VECTOR_DELIMITER` to smth else to overcome that. which also means vroom can't have comma-separated args to parse into a vector. or we change the separator for the `heuristic_params_arg`?

src/main.cpp
```diff
@@ -10,16 +10,15 @@ All rights reserved (see LICENSE).
...
-#else
-#include "getopt_win.h"
-#endif
 
 #if USE_LIBOSRM
 #include "osrm/exception.hpp"
 #endif
 
+// so we don't split heuristic param by comma
+#define CXXOPTS_VECTOR_DELIMITER ';'
```
nilsnolde, line 19: keep this or change `heuristic_params_arg` separator? or maybe another value instead of `;`?

jcoupey: Thanks for the explanation. Let's keep the default behavior (so remove this definition) and simply change our own internal separator:

```patch
diff --git a/src/main.cpp b/src/main.cpp
index bc9c6bd..f9ab8c7 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -15,8 +15,6 @@ All rights reserved (see LICENSE).
 #include "osrm/exception.hpp"
 #endif
 
-// so we don't split heuristic param by comma
-#define CXXOPTS_VECTOR_DELIMITER ';'
 #include "../include/cxxopts/include/cxxopts.hpp"
 
 #include "problems/vrp.h"
diff --git a/src/utils/helpers.h b/src/utils/helpers.h
index a2c1c98..3e64e11 100644
--- a/src/utils/helpers.h
+++ b/src/utils/helpers.h
@@ -108,7 +108,7 @@ inline void log_LS_operators(
 
 inline HeuristicParameters str_to_heuristic_param(const std::string& s) {
   // Split command-line string describing parameters.
-  constexpr char delimiter = ',';
+  constexpr char delimiter = ';';
   std::vector<std::string> tokens;
   std::string token;
   std::istringstream tokenStream(s);
```

 This means we should now use `-e "0;NONE;0.3"` instead of `-e "0,NONE,0.3"`, but as this is an (undocumented) debug option we're not breaking anything.

jcoupey:
There is yet another problem with multiple occurrences of the same flag. Say I have an instance with both `"profile": "car"` and `"profile": "bike"` vehicles and I have two OSRM instances running locally for car (port 5014) and bike (port 5001). Passing various ports used to work but now result in an error mentioning the default port value:

```
$vroom -i example_1.json -p car:5014 -p bike:5001
[Error] Failed to connect to 0.0.0.0:5000
{"code":3,"error":"Failed to connect to 0.0.0.0:5000"}
```

The same kind of odd thing happens with `-a`. Say I have an OSRM server running on XX.XX.XX.XX for car and locally for bike, then:

- `vroom -i example_1.json -a car:XX.XX.XX.XX -a bike:0.0.0.0` does not error but both requests are sent locally to the bike instance
- `vroom -i example_1.json -a bike:0.0.0.0 -a car:XX.XX.XX.XX` yields `{"code":2,"error":"Invalid profile: bike."}`

It looks like only the last flag occurrence is taken into account?

nilsnolde:
Ah right totally never played around with the „new“ multi profile implementation. That’s right, it’s only one taken into account. I’ll change that. Is there any other flag that can be specified multiple times?

jcoupey:
>Is there any other flag that can be specified multiple times?

No, only `-a`, `-p` and the already covered `-e`.

nilsnolde:
ok, that should've done it @jcoupey

jcoupey:
Thanks @nilsnolde for your work on this! I just made a couple adjustments on the clang-format side and took the opportunity for a bit of help rephrasing.

PR #606, opened 2021-11-02
Parse user-defined routes as initial solution

### LINKED ISSUE #376: Initial routes
Is there - or are there any plans - for the possibility to pass initial routes to vroom which it can continue to optimize?

We have some really big datasets which could be handy to cluster, divide, solve, merge, and then let vroom continue to optimize along the seams.

jcoupey:
## Issue

Fixes #376 

## Tasks

 - [x] Add heuristic value for taking user-defined routes into account
 - [x] Adjust heuristic parameters in `Input`
 - [x] Take into account new option in `VRP::solve`
 - [x] Populate routes from `Vehicle::steps` vectors
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/src/structures/typedefs.h
+++ b/src/structures/typedefs.h
@@ -11,6 +11,7 @@ All rights reserved (see LICENSE).
 */
 
 #include <array>
+#include <cassert>
 #include <chrono>
 #include <limits>
 #include <list>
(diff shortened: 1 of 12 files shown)


src/structures/typedefs.h
```diff
@@ -94,6 +95,12 @@ struct HeuristicParameters {
                                 float regret_coeff)
     : heuristic(heuristic), init(init), regret_coeff(regret_coeff) {
   }
+
+  // Only makes sense for user-defined initial routes.
+  constexpr HeuristicParameters(HEURISTIC heuristic)
+    : heuristic(heuristic), init(INIT::NONE), regret_coeff(0) {
+    assert(heuristic = HEURISTIC::INIT_ROUTES);
```
M393, line 102: I think this is intended to be a comparison.
```suggestion
    assert(heuristic == HEURISTIC::INIT_ROUTES);
```

jcoupey: Of course, good catch! By coincidence and due to the current usage of the ctor, this has no effect. But it may hide an unintended use of the ctor at some point in the future, which was the reason for having an assert in the first place...

@M393 do you fancy submitting this change as a simple PR?

M393: I opened #609 with this fix.

Issue #625, opened 2021-12-12
Adding `location_index` field in output where the `location` field is present


krashish8:
The `location_index` field is relevant whenever custom matrices are used. It plays a similar role as the `location` field when the routing engine is used. So, all the objects have both of these fields to be used depending on the situation. However, I see that in the output, only the `location` field is present (in the `unassigned` key, and the `step` object).

The user-defined matrices and the `location_index` type of fields were introduced in v1.2.0. Do we have any reasons for omitting the `location_index` type of field from the output, or is it because this field was not required then, so was not added?

I feel that it would be better to also add the `location_index` index field in the `unassigned` key, and the `step` object in the output. Firstly, it would look more consistent - both the `location` and `location_index` fields could be present in the output, depending on whether the custom matrices or the routing engine is used. Secondly, getting the location index of the task in the output might be useful to the users who are using custom matrices. Obviously, the end-users can get the location index of the task manually, but I think it would be easy to do it in the vroom itself without adding any extra complexity.

If this looks okay, I would be happy to submit a PR. :-)

jcoupey:
Technically, we provide a way to uniquely identify any task in output (step or unassigned) with `type` + `id` so users should be able to go back to their full description provided in input in the first place. In this sense, any additional information can be considered redundant. In order to avoid bloating the API, I've been trying to resist providing too much stuff in output "for convenience" and turned down such requests before, e.g. to feature initial constraints in output.

Now the first exception to this implicit rule is locations, which were included from the start for the sake of being able to simply plot a solution only from the output. I agree that not having `location_index` available in output while they are mandatory in input with custom matrices is not really consistent. So let's extend the `location` logic in output to `location_index`, of course only in the case where custom matrices are provided in input.

jcoupey:
@krashish8 if you submit a PR appending some stuff to unassigned objects, you may want to kill two birds with one stone by handling "description" too, see #403. ;-)

krashish8:
Actually, trying to kill "three" birds with one stone, where one bird is entirely different from the other two ;-)

Maybe it's not a good idea to kill different birds (issues) with a single stone (PR), but the changes were quite small, so I did. As a matter of fact, I'm getting rid of all the issues that I created here :-)

Please review: #627

Issue #617, opened 2021-11-26
Adding -fPIC flag in makefile?


krashish8:
**Problem**
When using libvroom as a dependency for other projects, building VROOM without `-fPIC` flag gives an error. With the `-fpic` or -fPIC flag, gcc creates an object file with position-independent code. This is usually necessary when creating a shared object.

Is it a good idea to add this flag to the makefile? Or are there some tradeoffs with using this flag, hence, the users must do this themselves depending on requirements?

**To Reproduce**
Building vrpRouting (pgRouting) with VROOM without the `fPIC` flag gives an error: https://github.com/krashish8/vrprouting/runs/4330973080?check_suite_focus=true#step:9:70

Adding this flag before building VROOM resolves this error:
https://github.com/krashish8/vrprouting/blob/c1839b715ef64d769a5b025a3d9c0e68e1633a9c/.github/workflows/ubuntu.yml#L65-L76

As of now, we were updating the makefile and adding this flag before building VROOM, as mentioned in the above GitHub Actions workflow file. We can continue doing this, so it's not a big issue, still any thoughts on this?

krashish8:
> ### Why is this a problem within pgRouting and not in our [toy example](https://github.com/VROOM-Project/vroom/tree/master/libvroom_examples)?

I'd say that pgRouting is a PostgreSQL extension, and a PostgreSQL extension is a shared library and not a standalone executable. We can't normally link a static library to a dynamic library, so we need to add -fPIC flag to make the code position independent, so that it can be linked to a dynamic library.

I'm not an expert on this, so I'll quote some links from StackOverflow answers that I referred :)

Quoting from ["When to use dynamic vs. static libraries"](https://stackoverflow.com/questions/140061/when-to-use-dynamic-vs-static-libraries):

> Static libraries increase the size of the code in your binary. They're always loaded and whatever version of the code you compiled with is the version of the code that will run.
> 
> Dynamic libraries are stored and versioned separately. It's possible for a version of the dynamic library to be loaded that wasn't the original one that shipped with your code if the update is considered binary compatible with the original version.
> 
> Additionally dynamic libraries aren't necessarily loaded -- they're usually loaded when first called -- and can be shared among components that use the same library (multiple data loads, one code load).

So, if we only want to link it into programs, it doesn't need PIC code. Otherwise, if we intend to link shared libraries against it, we need PIC code in the static library.

[...]

PR #627, opened 2021-12-17
Add `location_index` and `description` field to output, add shared target to makefile

### LINKED ISSUE #403: add 'description' to unassigned
First and foremost I'd like to thank you for all your work on vroom project!

Currently for `jobs` and `shipments` solutions, `steps` contains their `description`s -- this is very useful for programatically tracking their initial locations (e.g. we add our internal ID's there.)

However if the solution is not found, they are added to `unassigned` array without their corresponding `description`s.  This makes it harder for us to figure out their initial locations (have to indirectly go through initial json input).

I think it would be very useful if `unassigned` stops had their initial `description`s.  Also similarly for `start` and `end` steps (would be great if those had their vehicle `startDescription` and `endDescription`).  Thanks!

krashish8:
## Issue

Fixes #403, #617, #625

## Tasks

 - [x] Add `description` key to unallocated tasks in the output, if provided (#403)
 - [x] Add `location_index` key to to unassigned tasks and each step, if provided (#625)
 - [x] Add shared target to makefile (`make shared`), for creating Position Independent Code (#617)
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Added
+
+- `description` key to unallocated tasks in the output, if provided (#403)
+- `location_index` key to to unassigned tasks and each step, if provided (#625)
+- Shared target to makefile, for creating Position Independent Code (#617)
+
 ## [v1.11.0] - 2021-11-19
 
 ### Added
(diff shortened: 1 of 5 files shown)


krashish8:
It's funny how after adding the `description` and `location_index` keys, I realized that these changes only affect the end JSON-API users. The users that are using libvroom in C++ are not affected, and they could get this done even earlier. So, in truth, this change won't make any difference to us, but might be a bit helpful to the community! :-)

jcoupey (approved):

jcoupey:
Thanks for the PR!

> this change won't make any difference to us, but might be a bit helpful to the community!

You're definitely earning some open-source karma points here. ;-)

Issue #634, opened 2021-12-26
VROOM python wrapper


jonathf:
Hello there!

I've built a basic wrapper to VROOM in Python:
https://github.com/jonathf/pyvroom

I made it partially because I wanted to get more familiar with the VROOM code base, partially to improve on my own pybind11 skills, but also because I want to make some statistical experiments with vroom, and I'm likely much faster to do so in Python.

Currently I am using a fork of this repo where I have added some minimal patches to improve the Python interface. Preferably from my point of view, these patches should live upstream here in this repo. Would you guys be open to discussing improving the VROOMS interface to Python?

nilsnolde:
having done similar bindings in the past for related projects (mostly routing engines), gotta say: kudos! really clean and nice! had a similar goal in mind for early this year, to be able to easily use vroom in a QGIS plugin, pretty sure I'd have chosen the lazy way and just expose one high-level function accepting basically the JSON API's parameters :sweat_smile: 

just skimmed for 5 mins but learnt more than in days of suffering through python packaging myself (and some neat pybind11 code!).. did it work cross-platform already?

jonathf:
Thank you for the kind words. It helps a lot that the VROOM is clean and nice as well. :)

For this iteration I put the effort in on making Linux install work smooth with bundled binaries on Python 3.7-3.10. Getting from here to MacOS should be easy enough I think. I'm unsure how complicated it will be to get the builder to support Windows.

nilsnolde:
the manylinux images are definitely a godsend:) I'm really suprised though that you can easily build the vroom lib with the manylinux2014 tag, which [seems to have gcc4.8](https://pkgs.org/search/?q=gcc) and thus zero c++17 compatibility (and as I remember vroom makes pretty heavy use of c++17 features like `std::optional` etc).. I only built with the manylinux_x_y tag which runs debian 9 and gcc 6, mostly for other dependency versions but also bcs of gcc.

anyways, sorry for the spamming of this issue, fairly unrelated stuff.. if you're up for it, I'll gladly help with win integration and support of the bindings in general, I'll make use of them as well soon and I'd need win compatibility. we can discuss more in your repo.

jonathf:
Yes, indeed. I had to dig a bit, manylinux2014 is actually the only one variant of manylinux that fully supports c++17 currently:
https://github.com/pypa/manylinux/issues/1012

And thanks for the offer, it would be much appriciated. I'll tag you in the appropriate issue.

[...]

PR #637, opened 2021-12-31
remove braced-scalar-init

jonathf:
## Issue

#634, this issue prevents compiling on clang, which is what is used to compile pyvroom on macos.

## Tasks

 - [X] Update `CHANGELOG.md` (remove if irrelevant)
 - [ ] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Removed redundant braced-scalar-init in Matrix declaration (#637)
+
 ## [v1.11.0] - 2021-11-19
 
 ### Added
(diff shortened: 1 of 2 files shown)


jcoupey:
Thanks for spotting this and submitting a PR!

I'm curious which version of clang is raising an error? This has never been a problem before and we do have CI checks for [both GCC and clang](https://github.com/VROOM-Project/vroom/blob/master/.github/workflows/vroom.yml) (with varying versions over time).

jonathf:
I am sorry, looks I was too quick to post a PR and the real issue was my side of things. I was too quick and read one of the error messages I got from autometed MacOS-build, and interpreted it incorrectly.

Closing as PR is not relevant.

PR #638, opened 2021-12-31
Remove duplicate definition of LocalSearch

jonathf:
## Issue

#634:
Python requires a large list of includes in memory at all given times. Using the same name in the same namespace causes compile issues.

The proposed minimal solution is to just rename the conflicting variable, but removing the name all together is also possible here as it is only used for convinience.

## Tasks

 - [X] Update `CHANGELOG.md` (remove if irrelevant)
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Remove duplicate definition of LocalSearch (#638)
+
 ## [v1.11.0] - 2021-11-19
 
 ### Added
--- a/src/problems/cvrp/cvrp.cpp
+++ b/src/problems/cvrp/cvrp.cpp
@@ -35,6 +35,8 @@ namespace vroom {
 
 using RawSolution = std::vector<RawRoute>;
 
+namespace cvrp {
+
 using LocalSearch = ls::LocalSearch<RawRoute,
                                     cvrp::UnassignedExchange,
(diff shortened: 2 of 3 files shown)


jcoupey:
Thanks for the PR. If I get it right, the conflict is between the local `LocalSearch` type defined with "using" in `vrptw.cpp` and the one defined in `src/algorithms/local_search/local_search.h`? Then the same problem probably exists within `src/problems/cvrp/cvrp.cpp`?

What if we use a more specific local name, e.g. `using VRPTWLocalSearch = ...` and `using CVRPLocalSearch = ...` in `vrptw.cpp` and `cvrp.cpp` respectively? Or maybe even better we can namespace them as `vrptw::LocalSearch` and `cvrp::LocalSearch`?

jonathf:
Yes, that follows. pyvroom isn't feature complete yet, so it will likely not catch everything round one.

I like the namespace solution best. I can implement it if you like.

jonathf:
It is update now, @jcoupey.

I am realizing through the trenches of undeciphrable C++ error messages, that the name conlfict might not be against `src/algorithms/local_search/local_seach.h` since it is already in another namesspace. The conflict is more likely between `cvrp.cpp` and `vrptw.cpp` which both uses the root `vroom` namespace. Either way, butting both in seperate namespaces solves the issue on my side.

jcoupey (approved):

PR #647, opened 2022-01-09
Move priority check to Job constructor instead of input parser

### LINKED ISSUE #628: Move priority check to Job constructor instead of input parser
I see that the range of task priority is restricted between [0, 100] for the JSON-API user in the `input_parser.cpp`, but such checks are not provided in the internal C++ code. So, in C++, the range of priority can go high up to the upper limit of `uint64`.

https://github.com/VROOM-Project/vroom/blob/b06900ea8d06cbe8913edd0277850a0844978580/src/utils/input_parser.cpp#L112-L124

Shall we consider this as a "feature" instead of a bug, and assume that priority checks won't be added in the future in C++? ;-)

This is actually useful if we want to have some "pre-planned tasks" whose schedule won't change. Obviously, if someone is using libvroom, then custom checks can be added on the client-side for priority before sending data to VROOM.

So, on the client-side, one can keep the priority of such pre-planned tasks "very very high" (such as 10^9, instead of 100 because there is no restriction in libvroom), and let the other normal tasks have priority in the range [0, 100]. I understand that this still won't guarantee that pre-planned tasks are always allocated some schedule, but their probability would be very high. And since no checks for priority are added on the VROOM side, this feature can be exploited to make this happen. :-)

krashish8:
## Issue

Fixes #628 

## Tasks

 - [x] Move priority check to Job constructor instead of input parser
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/src/utils/helpers.h
+++ b/src/utils/helpers.h
@@ -305,6 +305,12 @@ inline void check_tws(const std::vector<TimeWindow>& tws) {
   }
 }
 
+inline void check_priority(const Priority priority) {
+  if (priority > MAX_PRIORITY) {
+    throw Exception(ERROR::INPUT, "Invalid priority value.");
+  }
+}
+
 inline Solution format_solution(const Input& input,
                                 const RawSolution& raw_routes) {
   std::vector<Route> routes;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Fixed
 
 - Remove duplicate definition of LocalSearch (#638)
+- Move priority check to Job constructor instead of input parser (#628)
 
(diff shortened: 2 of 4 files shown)


src/utils/helpers.h
```diff
@@ -305,6 +305,12 @@ inline void check_tws(const std::vector<TimeWindow>& tws) {
   }
 }
 
+inline void check_priority(const Priority priority) {
+  if (priority > MAX_PRIORITY) {
+    throw Exception(ERROR::INPUT, "Invalid priority value.");
```
jcoupey, line 310: Exceptions are changed simultaneously in #641, I'll adjust this myself in the merge commit.

jcoupey (approved):

Issue #651, opened 2022-01-13
Update libosrm version in CI


jcoupey:
We currently check builds against v5.25.0 of libosrm, but v5.26.0 has been released a while back. We should update to trigger caching a new libosrm version for the GitHub Actions workflow.

gauravagrwal:
Hi @jcoupey, I would to take this.
Just to confirm here we need to update the version of libosrm to v.5.26.0 in the workflow file ?

jcoupey:
Yes, then we'd have to make sure the new version is properly cached for subsequent workflow runs.

gauravagrwal:
@jcoupey, I believe the new version will be cached as for caching we are using osrm-tag which now has the updated version.

jcoupey:
Landed with #652.

PR #652, opened 2022-01-13
update libosrm version from v5.25.0 -> v5.26.0

gauravagrwal:
## Issue 
#651 

## Tasks

 - updated the libosrm version

### DIFF
--- a/.github/workflows/vroom_libosrm.yml
+++ b/.github/workflows/vroom_libosrm.yml
@@ -14,7 +14,7 @@ on:
       - '.github/workflows/libosrm.yml'
       - '**libosrm_wrapper**'
 env:
-  osrm-tag: v5.25.0
+  osrm-tag: v5.26.0
 jobs:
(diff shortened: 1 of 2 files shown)


jcoupey:
This is weird, only the vroom workflow [has been run](https://github.com/VROOM-Project/vroom/runs/4803640875?check_suite_focus=true), not vroom + libosrm. I would have expected it to run too since this PR touches a listed file:

https://github.com/VROOM-Project/vroom/blob/f3f433311a51d600b221a755c26792df2ab3bdce/.github/workflows/vroom_libosrm.yml#L10-L15

jcoupey:
You're right, I never noticed the name mismatch. This would be a typo from the time we switched to GitHub actions. Can you take the opportunity of this PR to rename the path to `vroom_libosrm.yml`?

jcoupey:
Thanks @gauravagrwal for the PR.

PR #659, opened 2022-01-25
Add local search operators stats

### LINKED ISSUE #658: Debug info on local search operators
For development/debugging, it would be useful to compute some stats on local search operators. Typically how many moves of each type are tested and how many are applied (maybe with associated gain). I don't think we have a sound way to decide how many of the tried moves are actually valid because most of them are usually discarded without even checking validity, based on the potential gain (or a lower bound of the gain).

jcoupey:
## Issue

Fixes #658

## Tasks

 - [x] Store type/number of moves tried at `LocalSearch` level
 - [x] Store type/number of moves applied at `LocalSearch` level
 - [x] Update `CHANGELOG.md`
 - [x] review


krypt-n:
Did you by any chance benchmark this? It seems to have quite the impact (at least 15% if not more) and to be enabled by default. I think we should recommend users of vroom to compile with `-DNDEBUG`

jcoupey:
No, I didn't. I suspect that this is related to using an `unordered_map` to store the counts based on operator name:

https://github.com/VROOM-Project/vroom/blob/master/src/algorithms/local_search/local_search.cpp#L548

While this is constant-time-ish, there are *a lot* of occurrences of those updates. Setting aside the fact that users are already advised to build against tagged versions (so with `-DNDEBUG` enabled), I agree we can do better. Options include:

- store counts in plain `std::array`s by assigning a rank to all operators (instead of a name), this would be a bit less straightforward to follow and maintain, especially for [the count of applied moves](https://github.com/VROOM-Project/vroom/blob/b8bb972d149df3d7aa992aa218bf6485e52b5b73/src/algorithms/local_search/local_search.cpp#L1427) which is the reason I went for name indexing in the first place;
- since this is only really required for dev purposes, put this behind another flag that one has to add explicitly.

We should probably do both.

krypt-n:
> Setting aside the fact that users are already advised to build against tagged versions (so with -DNDEBUG enabled),

Ah I wasn't aware of that. Then it is not as much of an issue as I thought.

> I suspect that this is related to using an unordered_map to store the counts based on operator name

String hashing appears to be the bottleneck according to my profiling. I think a quick fix could be to replace the string names with an enumeration 
```cpp
enum OperatorName {
    UnassignedExchange,
    CrossExchange,
    ...
}
```
and an additional array to convert the enum-values to strings for printing.

jcoupey:
@krypt-n I've sketched some changes in #667 based on your suggestion. Feel free to review it if you can!

Issue #665, opened 2022-01-29
Make cleaner separation between parser and main.


jonathf:
Currently the main interface creates CLArgs to parser arguments, which (depending on input) get passed down into the vroom machinery.
For the most part this is done cleanly in such a way that you don't need CLArgs to use any machinery, with input parser being the only exception.

I propose we make the interface cleaner by keeping CLArgs contain to the main file and just passing it's attribute along explicitly.

jcoupey:
Yes, makes perfect sense.

jonathf:
Before mergin though (or after for that matter), I am wondering if I need to make an adjustment here.

The problem I am realizing I will be faced with is that as soon as sub-class Input, using the parser will be unavailable.
So to be able to subclass Input, I either need to copy content over, or make my own parser.

A better solution, I think, is to initialize Input in `main` and only pass `input` and `input_str` to the parser function. That way, any subclass of Input can also use the parser to fill in content from json.

Sounds good?

jcoupey:
What would be the need behind inheriting from `Input`?

jonathf:
I subclass the C++ classes in Python to add lots of extra Python features which is a lot more difficult to do from pure C++.

jcoupey:
I have no problem to use another approach and maybe you can sketch that in a separate PR, but I fear that you won't be able to create an `Input` object outside of the parsing code since the ctor requires to know the size of the amounts arrays in the first place.

jcoupey:
This has been merged with #666, thanks @jonathf.

PR #666, opened 2022-01-29
Make cleaner separation between parser and main.

jonathf:
Currently the main interface creates CLArgs to parser arguments, which (depending on input) get passed down into the vroom machinery.
For the most part this is done cleanly in such a way that you don't need CLArgs to use any machinery, with input parser being the only exception.

I propose we make the interface cleaner by keeping CLArgs contain to the main file and just passing it's attribute along explicitly.

## Issue

#665

## Tasks

 - [X] Update `CHANGELOG.md`
 - [ ] review

### DIFF
--- a/src/structures/vroom/amount.h
+++ b/src/structures/vroom/amount.h
@@ -134,7 +134,7 @@ public:
     return elems.data();
   };
 #endif
-  
+
   template <class AmountExpression>
   Amount& operator+=(const AmountExpression& rhs) {
     assert(this->size() == rhs.size());


src/utils/input_parser.h
```diff
@@ -18,7 +18,7 @@ namespace io {
 
 struct CLArgs;
```
jcoupey, line 19: The `CLArgs` forward declaration is not needed any more.

src/utils/input_parser.h
```diff
@@ -18,7 +18,7 @@ namespace io {
...
 
-Input parse(const CLArgs& cl_args);
+Input parse(std::string input_str, Servers& servers, ROUTER router, bool geometry);
```
jcoupey, line 21: We should make `input_str` a `const std::string&` to avoid copying it. Also since we used to pass a `const CLArgs&`, we may as well have a `const Servers&`: `parse` should not change the server config and this is also what the `Input` ctor consumes down the line.

jcoupey (changes requested):
Simple change but a good readability improvement. I guess this is the benefit of having people look at the code from a "downstream" perspective! ;-)

src/utils/input_parser.h
jonathf: Not relevant any more as signature has changed.

jonathf:
I've pushed updated changes now.

To avoid the issue with requiring known amount_size, I set some asserts in place to assure that amount_size is always defined first. 

If you think this is too invasive, I will be happy to roll back to the first iteration (with suggestions) and continuing the conversation for a new PR in the #665.

jcoupey (changes requested):
This is clean. We're loosing constness of `Input::_amount_size` but I think it's fine since it's private and users have to do things right anyway via `set_amount_size`.

jonathf:
I've applied alll suggestions and rebased the branch to the tip of master.

jcoupey:
The PR shows up as closed since I messed a bit with the branch but it's merged all right. Looks like GitHub is confused as I squashed the last bugfixes commits into a single one and pushed the merge commit to `master` before force-pushing this branch.

PR #667, opened 2022-01-31
Operator stats enhancement

jcoupey:
## Issue

Follow-up to #659

## Tasks

 - [x] Add a dedicated build flag for local search operator stats
 - [x] Replace string names with an enum
 - [x] review


src/algorithms/local_search/local_search.h
```diff
@@ -65,10 +52,10 @@ class LocalSearch {
   std::vector<Route>& _best_sol;
   utils::SolutionIndicators _best_sol_indicators;
 
-#ifndef NDEBUG
+#ifdef LOG_LS_OPERATORS
   // Store operator usage stats.
-  std::unordered_map<std::string, unsigned> tried_moves;
-  std::unordered_map<std::string, unsigned> applied_moves;
+  std::unordered_map<OperatorName, unsigned> tried_moves;
+  std::unordered_map<OperatorName, unsigned> applied_moves;
```
krypt-n, line 58: Did some quick benchmarking on a single instance, and this still seems to have quite the performance impact.

If we use an `std::array` however:
```
std::array<unsigned, OperatorName::MAX> tried_moves;
std::array<unsigned, OperatorName::MAX> applied_moves;
```
compiling with or without `LOG_LS_OPERATORS` seems to have a negligible impact on performance.
This requires that we add an additional `MAX`-variant to the OperatorName enum, which must always be the last variant. Additionally the 0-initialization in the constructor should be replaced by `tried_moves.fill(0)` and `applied_moves.fill(0)`. The rest of the code should work as-is.

krypt-n (changes requested):

src/algorithms/local_search/local_search.h
jcoupey: Thanks for your comments! Of course the use of `unordered_map` is still a bottleneck. Switching to `array`s is the right move, which I figured later yesterday after pushing... I'll make the switch and initialization adjustments.

jcoupey:
This should look better, both codewise and for the computing times overhead.

krypt-n (approved):

Issue #675, opened 2022-02-18
Add support for compiling the project without ASIO


aardvarkk:
Currently, compilation is dependent upon ASIO libraries for making routing calls. If the only functionality required is for custom matrices (and no route geometry output is requested), then no routing support is required. This speeds up the compilation process and has reduces build dependencies.

PR #676, opened 2022-02-18
Add support for disabling use of routing entirely at the compiler level

aardvarkk:
- Allows the compile to take place without ASIO
- Optimization still works if using a custom cost matrix
- Default compilation behaviour should be identical to previous
- To compile, just run `USE_ROUTING=false make`

## Issue

Fixes https://github.com/VROOM-Project/vroom/issues/675

### DIFF
--- a/src/makefile
+++ b/src/makefile
@@ -31,15 +31,15 @@ ifeq ($(USE_ROUTING),false)
 	SRC := $(filter-out $(wildcard ./routing/*.cpp), $(SRC))
 else
 	LDLIBS += -lssl -lcrypto
-endif
 
-# Checking for libosrm
-ifeq ($(shell pkg-config --exists libosrm && echo 1),1)
-	LDLIBS += $(shell pkg-config --libs libosrm) -lboost_system -lboost_filesystem -lboost_iostreams -lboost_thread -lrt -ltbb
-	LIBOSRMFLAGS = $(shell pkg-config --cflags libosrm)
-	CXXFLAGS += $(filter-out -std=c++14, $(LIBOSRMFLAGS)) -D USE_LIBOSRM=true
-else
-	SRC := $(filter-out ./routing/libosrm_wrapper.cpp, $(SRC))
+	# Checking for libosrm
+	ifeq ($(shell pkg-config --exists libosrm && echo 1),1)
+		LDLIBS += $(shell pkg-config --libs libosrm) -lboost_system -lboost_filesystem -lboost_iostreams -lboost_thread -lrt -ltbb
+		LIBOSRMFLAGS = $(shell pkg-config --cflags libosrm)
+		CXXFLAGS += $(filter-out -std=c++14, $(LIBOSRMFLAGS)) -D USE_LIBOSRM=true
+	else
+		SRC := $(filter-out ./routing/libosrm_wrapper.cpp, $(SRC))
+	endif
 endif
 
 # Checking for libglpk based on whether the header file is found as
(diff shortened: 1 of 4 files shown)


jcoupey:
Thanks for submitting this idea and a PR. It makes perfect sense to have an easy way to drop unused stuff when only using custom matrices.

src/structures/vroom/input/input.cpp
```diff
@@ -57,6 +59,7 @@ void Input::set_geometry(bool geometry) {
   _geometry = geometry;
 }
 
+#if USE_ROUTING
 void Input::add_routing_wrapper(const std::string& profile) {
```
jcoupey, line 63: The `add_routing_wrapper` definition should also be conditioned by `USE_ROUTING`.

src/structures/vroom/input/input.cpp
```diff
@@ -645,6 +649,7 @@ void Input::set_matrices(unsigned nb_thread) {
   for (const auto& profile : _profiles) {
     thread_profiles[t_rank % nb_buckets].push_back(profile);
     ++t_rank;
+    #if USE_ROUTING
```
jcoupey, line 652: A plain removal of this portion is a bit extreme. As a result, we hit the assert on line 667 when running on a "usual" problem with no custom matrix:

```
vroom -i docs/example_1.json
vroom: structures/vroom/input/input.cpp:677: vroom::Input::set_matrices(unsigned int)::<lambda(const std::vector<std::__cxx11::basic_string<char> >&)>: Assertion `d_m != _durations_matrices.end()' failed.
Aborted (core dumped)
```

The expected behavior here would be to get a nice and explicit error. Maybe we could keep `add_routing_wrapper` after all but simply have it throw with "VROOM compiled without routing support" a bit like here:

https://github.com/VROOM-Project/vroom/blob/e3938092e60e1b2d2c5da58feef0b5fe819cd5f4/src/structures/vroom/input/input.cpp#L88

jcoupey (changes requested):
I think this is fine as soon as we have a proper handling of input with missing custom matrices.

Please also add a changelog entry and [run `clang-format`](https://github.com/VROOM-Project/vroom/blob/master/CONTRIBUTING.md#automatic-formatting).

src/makefile
```diff
@@ -25,6 +26,13 @@ SRC = $(wildcard *.cpp)\
...
 			$(wildcard ./utils/*.cpp)
 
+# Remove routing if we don't require it
+ifeq ($(USE_ROUTING),false)
+	SRC := $(filter-out $(wildcard ./routing/*.cpp), $(SRC))
+else
+	LDLIBS += -lssl -lcrypto
+endif
+
 # Checking for libosrm
```
jcoupey, line 36: Update: checking whether `libosrm` is installed should only happen with `USE_ROUTING=true`. Disabling routing will remove the libosrm includes anyway so we'd then be linking a lot of unnecessary stuff.

aardvarkk:
@jcoupey Thanks for the feedback. I believe I have incorporated all of the requested changes. Please let me know if you feel I've missed anything.

jcoupey:
Thanks for the changes, the diff is actually much simpler now. I'll simply remove modification for older changelog entries (probably automatic IDE formatting) before merging.

jcoupey:
Except for the entry on PDPTW, [unreleased stuff](https://github.com/VROOM-Project/vroom/blob/master/CHANGELOG.md#unreleased) is pretty much technical but not directly related to solving. The next release will probably wait for more solving-related stuff to be included in `master`.

PR #695, opened 2022-04-11
polylineencoder added as a submodule

### LINKED ISSUE #686: Submodule polylineencoder
We have a dependency to [polylineencoder](https://github.com/vahancho/polylineencoder/) currently taking the form of a header file at `include/polylineencoder.h`.

Since we now use git submodules for #602 it would be both easy and clean to replace our own copy of the file with a submodule.

erdemuysal:
## Issue

fixes #686 

I didn't update API & CHANGELOG because this PR shouldn't have changed anything in manner of output. Am I right?

## Tasks

 - [x] Add `polylineencoder` as a submodule
 - [x] Include `polylineencoder` from submodule instead of plain header
 - [x] Remove archaic plain header
 - [x] Point to [`polylineencoder`](https://github.com/vahancho/polylineencoder/releases)@[`v1.1.2`](https://github.com/vahancho/polylineencoder/tree/v1.1.2)
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey (changes requested):
Thanks for picking this up!

Nothing is required indeed on the API side but you're welcome to add an entry to the bottom of the current Unreleased/Changed section.

Nitpicking: I think we should target release tags for submodules, so instead of using latest `master` at f15223, could you point [to `v1.1.2`](https://github.com/vahancho/polylineencoder/releases)?

.gitmodules
```diff
@@ -1,3 +1,7 @@
 [submodule "include/cxxopts"]
 	path = include/cxxopts
 	url = https://github.com/jarro2783/cxxopts
+[submodule "include/polylineencoder"]
+	path = include/polylineencoder
+	url = https://github.com/vahancho/polylineencoder
+	branch = v1.1.2
```
jcoupey, line 7: I'm not familiar with the submodule options in git and a bit puzzled since there is no `v1.1.2` branch in `polylineencoder`, just a tag. Does specifying this will fallback to using the tag if no branch exists?

jcoupey (approved):

.gitmodules
erdemuysal: I have to say this sounds a bit confusing to me too. When you want to lock submodule to a spesific tag this trick becomes necessary AFAIK. Because when you run something like this.

```
git submodule update --init --remote
```

If branch isn't defined, submodule upgraded to latest commit on default branch and pollute tag-lock. When there is no release branch on the target repo, it isn't possible to determine the correct branch name. 

Since we have already locked the submodule to a tag, I chose to use the branch of the submodule to show its version for informational purposes.

This will only prevent you from updating submodules as above. When a submodule needs to be updated, it will be enough to run the following. Actually otherwise you should consider to not polluting submodule. 

```
git submodule set-branch --branch {desired_version} include/{module_name}
cd include/{module_name}
git fetch && git checkout {desired_version}
```

I don't think what I mentioned above is best-practice. Do you have any suggestions?

jcoupey: > Since we have already locked the submodule to a tag, I chose to use the branch of the submodule to show its version for informational purposes.

Right, I think I get it now. The actual commit hash we point to is not relying on this value but stored by git anyway and shows up on status:

```
$ git submodule status
 c74846a891b3cc3bfa992d588b1295f528d43039 include/cxxopts (v3.0.0)
 bea1070d8dd974d273f39f1cf2c4b393f5594f99 include/polylineencoder (1.0.0-29-gbea1070)
```

I think we can remove the `branch`  value from `.gitmodules` since this is somehow redundant, and not pointing to an actual branch anyway.

erdemuysal: That makes sense 👍

Issue #682, opened 2022-03-03
Missing SwapChoice options


jcoupey:
The SWAP* operator implemented back in #507 goes through a set of precomputed options that seem interesting cost-wise to exchange jobs between two routes (not necessarily in place). A [`SwapChoice`](https://github.com/VROOM-Project/vroom/blob/06af8717ec292a13ce2f7a9a541e41180b9c56da/src/algorithms/local_search/swap_star_utils.h#L69-L75) object embeds the move information along with the estimated gain.

For each pair of ranks in source and target route, `SwapChoice` candidates are stored in a set (starting [here](https://github.com/VROOM-Project/vroom/blob/06af8717ec292a13ce2f7a9a541e41180b9c56da/src/algorithms/local_search/swap_star_utils.h#L199)). The good part of using a set here is that the candidates are sorted by decreasing gain. So when evaluating feasibility we can stop at the first valid move and discard further checks.

Now the problem is that storing in a set will also prevent inserting different moves with the same gain. The result is that an invalid move inserted first could "shadow" a valid move with same gain, whose validity won't be checked at all.

This problem is probably only theoretical in the wild since the probability of having different moves with the exact same gain using real travel times in seconds is close to zero.  But I've confirmed that this actually happens on a regular basis with benchmarks (e.g. Solomon instances), probably due to the use of euclidean distances and grid-based points. In that case, we're missing some valid `SwapChoice` that were handy, simply because they were not considered at all since they show the same gain as another choice that turns out to be invalid.

PR #696, opened 2022-04-15
Use insertion ranks bounds for LS operator pruning

### LINKED ISSUE #583: Prune local search moves based on TW constraints
This is somehow similar to #509 but whereas the latter focuses on costs this could be seen as the "time" counterpart.

Consider a job `j` with TW `[a(j), b(j)]` and the task `t` at rank `i` in route `r` with TW `[a(t), b(t)]`. The situation of disjoints TW can be summed up as:
- if `b(j) < a(t) + service(t) + duration(t, j)` then `j` cannot be inserted in `r` after rank `i`;
- if `b(t) < a(j) + service(j) + duration(j, t)` then `j` cannot be inserted in `r` before rank `i + 1`.

This provides an embarrassingly simple way to discard a lot of local search move lookups. Contrary to #509 this would not require any tuning, just avoiding some unnecessary work (creating operator object, computing potential gain then ending up deciding it's invalid anyway). We'd only need to maintain min/max insertion ranks for all jobs into any route, which would not be expensive since only computed upon route modifications.

Of course this would not bring much in situations with loose TW constraints, but I'd expect a significant impact on instances with a lot of disjoints TW, e.g. all the usual VRPTW/PDPTW benchmarks.

jcoupey:
## Issue

Fixes #583

## Tasks

 - [ ] Store insertion rank bounds based on earliest/latest time for all jobs/routes
 - [ ] Store weak insertion rank bounds based on TW for all jobs/routes
 - [ ] Update relevant insertion ranks upon route changes
 - [ ] Prune `MixedExchange` moves
 - [ ] Prune `CrossExchange` moves
 - [ ] Prune `TwoOpt` moves
 - [ ] Prune `ReverseTwoOpt` moves
 - [ ] Prune `IntraRelocate` moves
 - [ ] Prune `IntraOrOpt` moves
 - [ ] Prune `IntraExchange` moves
 - [ ] Prune `IntraCrossExchange` moves
 - [ ] Prune `IntraMixedExchange` moves
 - [ ] Benchmark reduction in number of operator lookups
 - [ ] Benchmark computing time reduction
 - [ ] Update `CHANGELOG.md`
 - [ ] review


jcoupey:
Here are some comparisons on the usual VRPTW benchmark classes using `-x 5` where results are nearly always the same in quality (a very small number of solutions have changed due to different search paths after fixing #682 along the way).

### Average computing times

Computed for all instances in each benchmark class.

| Bench         | aa3a1ee | fe87ad8 | delta (%) |
| ------------- |--------:|--------:|----------:|
| Solomon       | 751     | 385     | -48.7     |
| Homberger 200 | 4094    | 1865    | -54.4     |
| Homberger 400 | 20873   | 9526    | -54.4     |
| Homberger 600 | 52193   | 23961   | -54.1     |

So it's a steady 2x speedup! :tada:  Of course we're talking about instances where the TW are rather tight and constraining so the pruning approach does make a bigger difference.

### LS operators lookups

Here is the total number of LS operators lookups across all benchmark instances.

| Bench         |      aa3a1ee |     fe87ad8 | delta (%) |
| ------------- |-------------:|------------:|----------:|
| Solomon       |   2362150254 |   938754682 |     -60.3 |
| Homberger 200 |  14479752836 |  5258489935 |     -63.7 |
| Homberger 400 |  76064030714 | 25687922018 |     -66.2 |
| Homberger 600 | 155454871416 | 60966284162 |     -60.8 |


The total reduction shadows differences related to TW tightness in the instances. For example the reduction in number of applied operator ranges from -22.9% to -86.0% across all Solomon instances.

Issue #683, opened 2022-03-03
Wrong index values without custom matrix


jcoupey:
When providing custom matrices, users have to describe the matching between locations in the instance and matrices columns/lines. This happens using `start_index` and `end_index` keys for vehicles and `location_index` for tasks.

Those keys do not make sense when no custom matrix is provided. In that case, location index values are decided on the fly depending of the order in which locations are added.

The problem is that it's still possible to pass `*_index` keys without a matrix, and then we'll later rely on those index values to retrieve durations/costs internally. Generally in a way that is inconsistent with how the actual matrices have been defined, leading to weird optimization choices.

In order to fix this, we should error whenever a custom index has been used but no matrix has been manually set.

jcoupey:
Landed in #698.

PR #698, opened 2022-04-21
Fix/issue#683

PattyDePuh:
## Issue [#683](https://github.com/VROOM-Project/vroom/issues/683)

Created an accessor for the `has_custom_location_index` variable inside `Input`.
Accessor is used at the end of `parse()`, if no custom matrizes where found inside JSON input.

### DIFF
--- a/src/utils/input_parser.cpp
+++ b/src/utils/input_parser.cpp
@@ -580,13 +580,13 @@ void parse(Input& input, const std::string& input_str, bool geometry) {
         }
       }
     }
-  } else if (json_input.HasMember("matrix")) {
+  } else {
     // Deprecated `matrix` key still interpreted as
     // `matrices.DEFAULT_PROFILE.duration` for retro-compatibility.
-    input.set_durations_matrix(DEFAULT_PROFILE,
-                                get_matrix<Duration>(json_input["matrix"]));
-  } else {
-    // SanityCheck for inappropriate index attributes takes place in Input::set_matrices()
+    if (json_input.HasMember("matrix")) {
+      input.set_durations_matrix(DEFAULT_PROFILE,
+                                 get_matrix<Duration>(json_input["matrix"]));
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


jcoupey (changes requested):
Thanks for submitting a PR!

The changes you made would work when parsing a json problem with `location_index` values and no matrix but it would still be possible to have inconsistent definitions when using the C++ API.
This is pretty much the same as in #628: we should enforce the checks at a lower level in the `Input` class so that they apply both to json-defined problems and uses of the C++ API.

A good place to do this is at the beginning of [`Input::set_matrices`](https://github.com/VROOM-Project/vroom/blob/39438be9e5b61d88a5608a71b369994fc212f49a/src/structures/vroom/input/input.cpp#L634) because this function is called after the problem definition, prior to solving. There is already a check for missing location index there.

PattyDePuh:
Thanks for Feedback.

src/utils/input_parser.cpp
```diff
@@ -580,13 +580,13 @@ void parse(Input& input, const std::string& input_str, bool geometry) {
...
+  } else if (json_input.HasMember("matrix")) {
     // Deprecated `matrix` key still interpreted as
     // `matrices.DEFAULT_PROFILE.duration` for retro-compatibility.
-    if (json_input.HasMember("matrix")) {
-      input.set_durations_matrix(DEFAULT_PROFILE,
-                                 get_matrix<Duration>(json_input["matrix"]));
-    }
+    input.set_durations_matrix(DEFAULT_PROFILE,
+                                get_matrix<Duration>(json_input["matrix"]));
+  } else {
```
jcoupey, line 588: Not really a fan of a logic branch that does nothing, and I don't think we really need to comment here on stuff that other parts of the code is responsible for. We can probably revert all changes to `input_parser.cpp`?

jcoupey (commented):
Thanks for the changes, the check in `Input::set_matrices` does the job! Can you please add a changelog entry under the unreleased/fixed section?

PattyDePuh:
Everything is cleaned up and ready to merge.

jcoupey (approved):

Issue #699, opened 2022-04-22
Compiler warning when compiling with clang++


kkarbowiak:
When I compile vroom using clang++ (command `make CXX=clang++`), the compiler outputs a warning:
```
structures/vroom/input/input.cpp:681:41: warning: braces around scalar initializer [-Wbraced-scalar-init]
            d_m->second = Matrix<Cost>({{0}});
                                        ^~~
```
I believe this is a valid warning as the `Matrix`'s constructor accepts a scalar and the braces are redundant.

For the record, I'm using clang version 13.0.1.

jcoupey:
I think this is a left-over from the time where matrices where vectors of vectors and the initializer list made sense. This was changed back in #491 and it looks like we introduced a regression back then.

Now `Matrix<Cost>({{0}})` behaves like `Matrix<Cost>(0)` which creates an empty matrix, not a one-element matrix with a zero. I confirmed the problem by creating a single-location problem and got an explicit `[Error] location_index exceeding matrix size for car profile.`.

jcoupey:
Fixed in #700

PR #700, opened 2022-04-22
Fix compilation warning

kkarbowiak:
## Issue 699

This PR fixes issue 699.

The code could be changed further to use the default constructor.

## Tasks

 - [ ] run tests
 - [ ] review

### DIFF
--- a/src/structures/vroom/input/input.cpp
+++ b/src/structures/vroom/input/input.cpp
@@ -678,7 +678,7 @@ void Input::set_matrices(unsigned nb_thread) {
           // Durations matrix not manually set so defined as empty
           // above.
           if (_locations.size() == 1) {
-            d_m->second = Matrix<Cost>({{0}});
+            d_m->second = Matrix<Cost>(0);
           } else {
             auto rw = std::find_if(_routing_wrappers.begin(),
                                    _routing_wrappers.end(),


src/structures/vroom/input/input.cpp
```diff
@@ -678,7 +678,7 @@ void Input::set_matrices(unsigned nb_thread) {
           // Durations matrix not manually set so defined as empty
           // above.
           if (_locations.size() == 1) {
-            d_m->second = Matrix<Cost>({{0}});
+            d_m->second = Matrix<Cost>(0);
```
jcoupey, line 681: We should create a one-element matrix with a zero here, not an empty matrix, see https://github.com/VROOM-Project/vroom/issues/699#issuecomment-1106822394

jcoupey (changes requested):

src/structures/vroom/input/input.cpp
kkarbowiak: Alright. I've updated the code.

src/structures/vroom/input/input.cpp
```diff
@@ -678,7 +678,7 @@ void Input::set_matrices(unsigned nb_thread) {
           // Durations matrix not manually set so defined as empty
           // above.
           if (_locations.size() == 1) {
-            d_m->second = Matrix<Cost>({{0}});
+            d_m->second = Matrix<Cost>(1);
```
jcoupey, line 681: Here the [matrix ctor](https://github.com/VROOM-Project/vroom/blob/39438be9e5b61d88a5608a71b369994fc212f49a/src/structures/generic/matrix.cpp#L14-L16) will call `data.resize(1)` so that the underlying storage vector will have the right size. But I think we still need to initialize the single entry with a zero value.

kkarbowiak: `Cost` is an alias of `uint32_t` so there is no ctor to call. However, according to [reference](https://en.cppreference.com/w/cpp/container/vector/resize) in this overload of `std::vector::resize`, the items are default-inserted, which in this case boils down to zero-initialization.

There is no need to assign zero value explicitly, but I can add it for clarity. Please let me know what you think.

jcoupey: Thanks for the pointer, I didn't know that. I guess this is fine as is, thanks!

jcoupey (approved):

Issue #701, opened 2022-04-22
Assignment in assertions


kkarbowiak:
It looks like there are two cases of performing an assignment instead of equality check inside assertions:
* algorithms/validation/choose_ETA.cpp:931:
```c++
assert(current_delta_rank = nb_var + 1);
```
* algorithms/validation/choose_ETA.cpp:999:
```c++
assert(current_delta_rank = nb_var + 1);
```
I stumbled upon those thanks to compiler warnigs (compilation using g++ 11.2.0):
```
algorithms/validation/choose_ETA.cpp: In function ‘vroom::Route vroom::validation::choose_ETA(const vroom::Input&, unsigned int, const std::vector<vroom::VehicleStep>&)’:
algorithms/validation/choose_ETA.cpp:931:29: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
  931 |   assert(current_delta_rank = nb_var + 1);
      |          ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
algorithms/validation/choose_ETA.cpp:999:29: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
  999 |   assert(current_delta_rank = nb_var + 1);
      |          ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
```
Both `current_delta_rank` and `nb_var` are of type `unsigned`, and the assertions can only fail in case of wrap-around. Is this intended? I guess not, if only for the side effects of the assignment missing if compiling with `NDEBUG`.

I can send a PR that changes assignment to equality check, but I'm afraid I have no means to test this properly.

jcoupey:
Good catch, those are indeed intended to be regular equality checks, not assignments. Please submit a PR for this if you can, I'll handle the tests.

jcoupey:
Fixed in #702

PR #702, opened 2022-04-23
Fix assertions

kkarbowiak:
## Issue #701

This PR fixes the assertions by changing the assignments to equality checks as was originally intended.

## Tasks

 - [ ] test
 - [ ] review


jcoupey (approved):
I ran this on some plan mode test cases I used during dev and the new assert checks don't break. It simply means that the assertion where rightly placed at the time but ineffective.

Could you please add a changelog entry for this?

Issue #715, opened 2022-05-19
Change types to uint64


nurikk:
Hi, there
i've noticed that in typedef
https://github.com/VROOM-Project/vroom/blob/c238bf6d0590ea384bcc9a3aa84ca9f4fdde1099/src/structures/typedefs.h#L38

capacity is defined as `int64_t` (signed) but in input_parser 
https://github.com/VROOM-Project/vroom/blob/c238bf6d0590ea384bcc9a3aa84ca9f4fdde1099/src/utils/input_parser.cpp#L70

It's retrieved as `GetUint` unsigned

Additionally I changed Cost/Distance types to uint64 to enable support for larger problems instances with tasks on greater distance

jcoupey:
I just commented [the ongoing discussion](https://github.com/VROOM-Project/vroom/pull/716#pullrequestreview-978252552) on the `Capacity` side.

Regarding `Cost` and `Distance`, the only situation where I've had to switch to `uint64_t` is for a specific huge instance from TSPLIB. I'm not really fond of raising the size for some specific situation if it add a computational burden for "normal" problems. Did you somehow benchmark switching from `uint32_t` to `uint64_t`?

nurikk:
> Thanks for opening a dedicated ticket! I just commented [the ongoing discussion](https://github.com/VROOM-Project/vroom/pull/716#pullrequestreview-978252552) on the `Capacity` side.
> 
> Regarding `Cost` and `Distance`, the only situation where I've had to switch to `uint64_t` is for a specific huge instance from TSPLIB. I'm not really fond of raising the size for some specific situation if it add a computational burden for "normal" problems. Did you somehow benchmark switching from `uint32_t` to `uint64_t`?

Can I use this scrip for benchmark?
https://github.com/VROOM-Project/vroom-scripts/blob/master/benchmarks/compare_to_BKS.py

jcoupey:
> Can I use this scrip for benchmark?

Yes, though you don't necessarily need to run it yourself, this script is called from the `benchmarks/*/run.sh` scripts.

jcoupey:
Closing as this was discussed in #716, now closed.

PR #716, opened 2022-05-19
Change variable types to uint64_t

nurikk:
## Issue

[Change types to uint64](https://github.com/VROOM-Project/vroom/issues/715)

## Tasks

 - [ ] ...
 - [ ] review

### DIFF
--- a/src/structures/typedefs.h
+++ b/src/structures/typedefs.h
@@ -30,12 +30,12 @@ namespace vroom {
 // To easily differentiate variable types.
 using Id = uint64_t;
 using Index = uint16_t;
-using Cost = uint32_t;
+using Cost = uint64_t;
 using Gain = int64_t;
-using Distance = uint32_t;
-using Duration = uint32_t;
+using Distance = uint64_t;
+using Duration = uint64_t;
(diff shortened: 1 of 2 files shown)


src/structures/typedefs.h
```diff
@@ -30,12 +30,12 @@ namespace vroom {
...
-using Cost = uint32_t;
+using Cost = uint64_t;
 using Gain = int64_t;
-using Distance = uint32_t;
-using Duration = uint32_t;
+using Distance = uint64_t;
+using Duration = uint64_t;
 using Coordinate = double;
-using Capacity = int64_t;
+using Capacity = uint64_t;
```
krypt-n, line 38: Thanks for the PR.
This change seems dangerous at first glance. Amounts are basically `std::vector<Capacity>` and they are subtracted occasionally (e.g. https://github.com/VROOM-Project/vroom/blob/c238bf6d0590ea384bcc9a3aa84ca9f4fdde1099/src/structures/vroom/raw_route.cpp#L203). It may very well be that all subtractions guarantee that the result is positive, but this change should be tested nonetheless

Issue #728, opened 2022-06-21
Unit for Timeout is not obvious


kkarbowiak:
Currently, `Timeout` is defined as `std::optional<unsigned>` which is confusing, as it does not convey what the timeout unit is. Only by digging deeper into the code one finds a comment that internally milliseconds are used.

We can make it more self-explanatory and type-safe by changing the type to `std::optional<std::chrono::milliseconds>`.

jcoupey:
Definitely makes sense, thanks for the suggestion and the associated PR.

PR #729, opened 2022-06-21
Use `std::chrono::milliseconds` for Timeout value

kkarbowiak:
## Issue #728 

This PR changes the type of `Timeout` to `std::optional<std::chrono::milliseconds>` making it more type-safe and the unit used obvious.

This is an enhancement mostly for those who use VROOM as a library.

## Tasks

 - [ ] review

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -97,7 +97,10 @@ int main(int argc, char** argv) {
     try {
       if (!limit_arg.empty()) {
         // Internally timeout is in milliseconds.
-        cl_args.timeout = 1000 * std::stof(limit_arg);
+        const auto timeout = std::chrono::duration<double, std::milli>(
+          1000 * std::stof(limit_arg));
+        cl_args.timeout =
+          std::chrono::duration_cast<std::chrono::milliseconds>(timeout);
       }
     } catch (const std::exception& e) {
       throw cxxopts::OptionException("Argument '" + limit_arg +
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -98,7 +98,7 @@ int main(int argc, char** argv) {
       if (!limit_arg.empty()) {
         // Internally timeout is in milliseconds.
         cl_args.timeout = std::chrono::milliseconds(
-          static_cast<int>(1000 * std::stof(limit_arg)));
+          static_cast<std::chrono::milliseconds::rep>(1000 * std::stof(limit_arg)));
       }
     } catch (const std::exception& e) {
       throw cxxopts::OptionException("Argument '" + limit_arg +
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -97,7 +97,10 @@ int main(int argc, char** argv) {
     try {
       if (!limit_arg.empty()) {
         // Internally timeout is in milliseconds.
-        cl_args.timeout = 1000 * std::stof(limit_arg);
+        const auto timeout = std::chrono::duration<double, std::milli>(
+          1000 * std::stof(limit_arg));
+        cl_args.timeout =
+          std::chrono::duration_cast<std::chrono::milliseconds>(timeout);
```
jcoupey, line 103: I'm not sure why an intermediate variable is necessary here. Wouldn't  `cl_args.timeout = std::chrono::milliseconds(1000 * std::stof(limit_arg))` do the trick in the same way?

jcoupey (commented):
Thanks for the PR, this will make the timeout code more explicit.

src/main.cpp
kkarbowiak: `std::chrono` library is very cautious not to lose precision during conversions (at least unless explicitly asked to), so `std::chrono::milliseconds` cannot be constructed from a floating point value. This could be dealt with using a cast
`cl_args.timeout = std::chrono::milliseconds(static_cast<int>(1000 * std::stof(limit_arg)))` and I can change the PR if you prefer this version.

kkarbowiak: Done.

src/main.cpp
```diff
@@ -97,7 +97,8 @@ int main(int argc, char** argv) {
     try {
       if (!limit_arg.empty()) {
         // Internally timeout is in milliseconds.
-        cl_args.timeout = 1000 * std::stof(limit_arg);
+        cl_args.timeout = std::chrono::milliseconds(
+          static_cast<int>(1000 * std::stof(limit_arg)));
```
jcoupey, line 101: Thanks for the change. Something else occured to me: the cast here is limiting the max number of milliseconds one can use as a timeout to the max value of an `int`. This mean the max possible value is halved compared to the previous setup with an implicit cast to `unsigned`.

My understanding is that the actual limitation will depend on the number of bits used for `int` on a given platform. The standard says "at least 16" which is not much for a number of milliseconds. On the other hand, `std::chrono::milliseconds` is a helper type for `std::chrono::duration</*signed integer type of at least 45 bits*/, std::milli>`...

kkarbowiak: Cool. Switched to `std::chrono::milliseconds::rep`.

jcoupey:
Thanks @kkarbowiak for your suggestions and patience!

Issue #730, opened 2022-06-21
Longitude, Latitude order in Coordinates is not obvious


kkarbowiak:
`Coordinates` is defined as `std::array<Coordinate, 2>`, which does not inform on the order of the coordinate pair. Is it (Lat, Lon) or (Lon, Lat)? At first I made a mistake and provided the coordinates in incorrect order.

The situation could be improved by adding e.g. an enum: `enum { LON_INDEX = 0, LAT_INDEX = 1};`, or better yet, by defining coordinates as a struct with named members.

jcoupey:
Landed with #731.

PR #731, opened 2022-06-21
Use struct for Coordinates

kkarbowiak:
## Issue #730

This PR changes type of `Coordinates` to a struct with named members. This way it is more self-explanatory and less error-prone, as it makes it obvious on the order of coordinates used.

This is an enhancement mostly for those who use VROOM as a library.

## Tasks

 - [x] review


jcoupey (approved):
Thanks for the PR!

azmathmoosa:
# Note for people still wondering (like myself)

Coordinate format is longitude, latitude.

jcoupey:
> Coordinate format is longitude, latitude.

That's a classic source of error indeed. Also the reason why it's the very first information displayed on the [documentation page](https://github.com/VROOM-Project/vroom/blob/master/docs/API.md). ;-)

PR #752, opened 2022-08-01
Use Eval struct over plain Cost/Gain types

### LINKED ISSUE #738: Store several metrics in costs / gains
The current state of the codebase is that `Cost` values default to the `Duration` values as seen from each vehicle. This is not something we want to rely on since:
- we may want to allow more tuning for costs in the future;
- it's already possible to pass in arbitrarily different costs/durations using custom matrices.

Since #555, the separation of cost and duration is a thing in the codebase, even when values are equal. The problem is that when storing costs and gains (e.g. for local search operators), we're loosing information on durations. This information would be required for example in #273: on top of deciding whether a move is actually interesting, we'd need to assert it's valid for max travel time based on how much duration we're adding or removing.

I think the best way to add this information in an extensible way (looking at you #354) would be to change `Cost` and `Gain` to structs holding several metrics. For now we'd need the current `Cost` and `Duration` values; additional operators doing the math on both metrics; a comparison operator using only actual cost. The good part would be that we would compute all metrics at once with no need to change the client code for all computations within the heuristics and local search operators.

jcoupey:
## Issue

Fixes #738

## Tasks

 - [x] Create data structure to replace `Cost` and `Gain` values
 - [x] Adjust helper functions
 - [x] Adjust heuristics client code
 - [x] Adjust operators client code
 - [x] Adjust local search code
 - [x] Benchmark timing impact
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
I've run this PR against the usual benchmarks across various exploration levels.

### Solution obtained

All solutions are strictly identical with current `master` and this PR so it looks like I did not break anything with all the adjustments to switch from `Cost/Gain` to the new `Eval` struct.

### Timing

Computing time increase is in the +15 to 20% ballpark for CVRP and VRPTW instances and around +5 to +10% for PDPTW instances. So this will somehow eat a bit of the performance boost that was introduced in the latest release, but there is not really any other way around that I'm afraid.

### Possible improvement

The new `Eval` struct holds both `cost` and `duration` metrics and all the math operations are overloaded to operate on both members. So the computing overhead is due to switching from doing cost/gain math with basic types (`uint32_t` and `int64_t`) and a single value to doing it with this struct. I've implemented the various operators in a straightforward way but maybe it's possible to do a bit of compile-time magic to improve the efficiency here. Any input on that is welcome!

PR #755, opened 2022-08-02
Refactor SolutionIndicators struct

### LINKED ISSUE #750: Simplify SolutionIndicators logic
We use various heuristic tunings to reach to different initial solutions, then we apply our local search process on each and pick the best result in the end. The problem is that whenever two different tunings yield the same solution after the heuristic, we still run the local search on both instances. Thus we're potentially doing a significant amount of extra work for nothing.

In practice, heuristic + local search happen in a lambda that is called [across multiple threads](https://github.com/VROOM-Project/vroom/blob/master/src/problems/cvrp/cvrp.cpp#L251-L257), so we'd have to keep this setup for heuristic only, then filter out identical heuristic solutions after joining all threads, then parallelize the local search using another lambda.

jcoupey:
## Issue

Fixes #750.

## Tasks

 - [x] Dedicated function to compute `SolutionIndicators`
 - [x] Compare solutions beyond global indicators/cost
~~- [x] Split the solving multithreading between heuristic and local search~~
~~- [x] Remove "identical" solutions~~
 - [x] Update `CHANGELOG.md`
 - [ ] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Use `std::chrono::milliseconds` for `Timeout` value (#728)
 - Use `struct` for storing `Coordinates` instead of an `std::array` (#730)
+- Do not duplicate local search for identical heuristic solutions (#750)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 - Use `std::chrono::milliseconds` for `Timeout` value (#728)
 - Use `struct` for storing `Coordinates` instead of an `std::array` (#730)
-- Do not duplicate local search for identical heuristic solutions (#750)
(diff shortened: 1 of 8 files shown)


jcoupey:
I've tried benchmarking the changes here and the difference is not noticeable because:

- instances for which there are several similar heuristic solutions are mostly the easy/small ones so convergence is already super fast on those;
- medium to big instances seldom have similar heuristic solutions, even with `-x 5`;
- removing a couple local searches does not usually impact overall computing time in a context where multiple threads are used.

So I don't think it's actually worth adding this as is since:

- it adds quite some boilerplate to the solving code in `cvrp.cpp` and `vrptw.cpp`, splitting the multi-threading part;
- the current solution comparison based on global indicators can also "shadow" some solutions that look identical but for which actual routes are different; in some cases this actually cuts unwanted searches, sometimes leading to a slightly poorer overall solution.

Issue #764, opened 2022-08-25
Include response in the exception of an invalid routing response


SimonBradley1993:
Recently my team had issues with our optimisation process using Vroom and OSRM. After some investigation it turned out that our infrastructure was timing out a request that Vroom makes to OSRM - which took longer than the default 15 seconds.

To identify this root cause we modified Vroom to include the actual response that was received. 

It would be really great to have this change included so we can continue to use the tagged version of Vroom.

The changes that were needed were in `http_wrapper.cpp` with the following code block : 

```
// Removing headers.
  auto start = response.find("{");
  if (start == std::string::npos) {
    throw RoutingException("Invalid routing response: " + response);
  }
  auto end = response.rfind("}");
  if (end == std::string::npos) {
    throw RoutingException("Invalid routing response: " + response);
  }
```

jcoupey:
Done in #765.

PR #765, opened 2022-08-25
Include response in exception for invalid routing response

SimonBradley1993:
## Issue

Issue : https://github.com/VROOM-Project/vroom/issues/764

This includes the response in the Routing Exception to allow clients of the application to identify what the issue is.

## Tasks

 - [ ] ...
 - [ ] review

### DIFF
--- a/src/routing/http_wrapper.cpp
+++ b/src/routing/http_wrapper.cpp
@@ -70,11 +70,11 @@ std::string HttpWrapper::send_then_receive(const std::string& query) const {
   // Removing headers.
   auto start = response.find("{");
   if (start == std::string::npos) {
-    throw RoutingException("Invalid routing response.");
+    throw RoutingException("Invalid routing response: " + response);
   }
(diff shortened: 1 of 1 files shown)


jcoupey:
This makes sense, thanks for the PR!

While looking at the code I was surprised the same change was not necessary in `ssl_send_then_receive` and it turns out we're missing a couple `throw` there. Could you add them in this PR to fix #770 along the way?

Please also add a changelog entry.

SimonBradley1993:
> This makes sense, thanks for the PR!
> 
> While looking at the code I was surprised the same change was not necessary in `ssl_send_then_receive` and it turns out we're missing a couple `throw` there. Could you add them in this PR to fix #770 along the way?
> 
> Please also add a changelog entry.

Thanks for looking at this so quickly. I've made the changes you asked for. 

Let me know if there's anything else needed for this 🙂

jcoupey (approved):

Issue #776, opened 2022-09-06
Summary recreates zero amounts from amount size


kkarbowiak:
Currently one of `Summary`'s constructors accepts amount size, while it could accept the zero amount directly. This way it would have to be concerned less with the details of `Amount` such as its dimensionality.

jcoupey:
Makes sense. :+1:

jcoupey:
Changed in #777.

PR #777, opened 2022-09-06
Pass zero amount directly instead of its size

kkarbowiak:
## Issue #776 

This is a minor code improvement.

## Tasks

 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/src/utils/helpers.h
+++ b/src/utils/helpers.h
@@ -515,7 +515,7 @@ inline Solution format_solution(const Input& input,
                  [&](auto j) { return input.jobs[j]; });
 
   return Solution(0,
-                  input.zero_amount().size(),
+                  input.zero_amount(),
                   std::move(routes),
                   std::move(unassigned_jobs));
 }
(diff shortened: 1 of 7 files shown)


src/utils/helpers.h
```diff
@@ -515,7 +515,7 @@ inline Solution format_solution(const Input& input,
                  [&](auto j) { return input.jobs[j]; });
 
   return Solution(0,
-                  input.zero_amount().size(),
+                  input.zero_amount(),
```
jcoupey, line 518: A similar change is missing below in `format_solution` for TW routes (L928). I'm actually a bit surprised that passing an `unsigned` as `const Amount&` over there even compiles.

jcoupey (changes requested):

src/utils/helpers.h
kkarbowiak: Thank you for the review.

Yes, I realised I can do some more changes and will be back with them.

kkarbowiak: > I'm actually a bit surprised that passing an `unsigned` as `const Amount&` over there even compiles.

That I think is due to non-explicit `Amount(std::size_t size)` ctor allowing implicit conversion.

jcoupey (approved):

PR #782, opened 2022-09-09
Change `Input` to accept the zero amount instead of just its size

kkarbowiak:
## Tasks

 - [ ] review

### DIFF
--- a/libvroom_examples/libvroom.cpp
+++ b/libvroom_examples/libvroom.cpp
@@ -97,7 +97,7 @@ void run_example_with_osrm() {
 
   // Create one-dimension capacity restrictions to model the situation
   // where one vehicle can handle 4 jobs with deliveries.
-  problem_instance.set_amount_size(amount_dimension);
+  problem_instance.set_zero_amount(vroom::Amount(amount_dimension));
   vroom::Amount vehicle_capacity(1);
 
   vroom::TimeWindow vehicle_tw(28800, 43200); // Working hours.
--- a/src/utils/input_parser.cpp
+++ b/src/utils/input_parser.cpp
@@ -51,19 +51,18 @@ inline double get_double(const rapidjson::Value& object, const char* key) {
 
 inline Amount get_amount(const rapidjson::Value& object,
                          const char* key,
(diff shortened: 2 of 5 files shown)


libvroom_examples/libvroom.cpp
```diff
@@ -97,7 +97,7 @@ void run_example_with_osrm() {
 
   // Create one-dimension capacity restrictions to model the situation
   // where one vehicle can handle 4 jobs with deliveries.
-  problem_instance.set_amount_size(amount_dimension);
+  problem_instance.set_zero_amount(vroom::Amount(amount_dimension));
```
jcoupey, line 100: I see several drawbacks for changing from `Input::set_amount_size` to `Input::set_zero_amount`:
- in this example, client code is more complex, and in not way "oblivious to the fact that the amount has some dimensions";
- instead of just dealing with dimension (which is something the client code should know as part of the model) this requires to know how to build a zero `Amount` from its dimension which we could consider internals;
- this is dangerous since then any caller could initiate the zero member in `Input` with any non-zero amount.

src/utils/input_parser.cpp
```diff
@@ -51,19 +51,18 @@ inline double get_double(const rapidjson::Value& object, const char* key) {
 
 inline Amount get_amount(const rapidjson::Value& object,
                          const char* key,
-                         unsigned size) {
-  // Default to zero amount with provided size.
-  Amount amount(size);
+                         const Amount& zero_amount) {
+  Amount amount(zero_amount);
```
jcoupey, line 55: This init is probably simpler then, but I don't think this part of (internal) code can or even should be made non-aware of amount dimensions, since we have to check dimension consistency right below.

jcoupey (commented):
See comments for some concerns. Maybe we could keep the part of the changes that use only `Input::_zero` as reference and remove the `_amount_size` member, while sticking to passing around dimensions instead of `const Amount&`?

kkarbowiak:
Thank you for the review and comments!

I think I rushed a bit with this PR. Let me take a step back and think this over.

PR #788, opened 2022-09-21
Better costs modelling at vehicle level

### LINKED ISSUE #528: Fixed cost for a vehicle
## Current limitation

We lack a way to better model costs related to using different vehicles. Examples of (potentially) unexpected behavior that can be noticed in output for the sake of reducing overall travel time:

- using more vehicles (#495)
- using more expensive vehicles (#359)

The latter is also now probably more bound to happen as we now have multi-profile in place.

## API addition ideas

A way to pass this in the API would be to add a `cost` object at `vehicle` level holding a `v.cost.fixed` value. Then we'd have to provide a way to homogenize with current travel-related costs, maybe with a scaling coefficient like `v.cost.per_travelled_hour` or similar.

## Solving approach

This would require to go through all situations where we modify solutions during the local search phase (all neighboring operators, job removals/additions) to account for change in cost.

The fun part here is that we'd then be able to play with many new HVRP instances.

### LINKED ISSUE #732: Cost per hour in vehicle profile
Hello,
We need to use some sort of modeling that takes a cost per hour / vehicle. 
As far as I've seen from the documentation / discussions on issues, a custom cost matrix could be used for this purpose.
But I cannot understand from the documentation how to define such a matrix. 
I got that we use the first dimension of the matrix as an index for a vehicle definition, but cannot figure out what should be used as the other dimension. And above all I cannot figure out how to specify that the cost is a function of travel time.
Is there a way to model a custom cost matrix that is a function of travel time?

Thank you,
Marcel

jcoupey:
## Issue

This PR aims at improving the way costs can be fine-tuned at vehicle level, including:
- fixed costs (fixes #528);
- cost per time unit per vehicle (fixes #732).

## Tasks

 - [x] Parse fixed cost in input
 - [x] Parse time cost factor in input
 - [x] Update cost reporting in output
 - [x] Adjust `OrOpt` operator to account for fixed costs changes
 - [x] Adjust `RouteExchange` operator to account for fixed costs changes
 - [x] Adjust `PdShift` operator to account for fixed costs changes
 - [x] Adjust `Relocate` operator to account for fixed costs changes
 - [x] Adjust `ReverseTwoOpt` operator to account for fixed costs changes
 - [x] Tune vehicle ordering in heuristics
 - [x] Benchmark
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
Here is a quick summary of the results I'm getting with the state of this PR on the [HVRP benchmarks](https://github.com/VROOM-Project/vroom-scripts/tree/master/benchmarks/HVRP) I gathered.

`x` | Average gap (%) | Median gap (%) | Worst gap (%) | Average computing time (ms)
--: | --: | --: | --: | --:
0 | 7.41 | 6.35 | 44.14 | 72
1 | 5.66 | 4.73 | 44.14 | 168
2 | 4.81 | 4.14 | 44.14 | 362
3 | 4.35 | 3.85 | 43.6 | 603
4 | 3.96 | 3.45 | 43.6 | 1095
5 | 3.74 | 3.32 | 43.6 | 1730

Comments:

1. The gaps are quite hard to get down, those results have involved quite a lot of testing and introducing a new operator. The HVRP is significantly harder than the CVRP since there is an additional affectation problem to decide which vehicles to use. I've tested many things around this in the scope of our solving framework but there is no simple rule that works across all instances to decide on vehicles to pick based on fixed/variable cost and capacities.
2. I feel this is good enough for now because:
- as usual the real-life instances are probably much simpler usually (especially with a more or less adjusted fleet, the hard part with the benchmarks being that they have an "open-bar" setup on a lot of different vehicle types);
- computing times are still quite low (this is not even a release build) which leaves us with some room for further improvements later on.

Issue #786, opened 2022-09-17
Break only possible with empty vehicle


fbaierl:
### Feature

I would like to request a feature to only schedule breaks at a time, where the respective vehicle is empty (load is empty). 

### Motivation

We work with public transport so we transport people. For obvious reasons, we cannot schedule a break at a time, where customers are still inside the vehicle, so this would be an essential feature for all use-cases, where people are transported (with time-windows). 

### API

For the API, this could be achieved by setting an additional parameter (e.g. `while_transport`) to the break input definition:

| Key         | Description |
| ----------- | ----------- |
| `id` | integer |
| `time_windows` | an array of `time_window` objects describing valid slots for break start |
| [`while_transport`] | a boolean, that is true if the break can happen while transporting load  |
| [`service`] | break duration (defaults to 0) |
| [`description`] | a string describing this break |

### Implementation

As we previously discussed [here](https://github.com/VROOM-Project/vroom/issues/679), this might be possible with adapting 
`TWRoute::is_valid_addition_for_tw`.

### Support

We would be willing to fund the implementation of this feature. Would it be possible to get a quote for this?

jcoupey:
I've never encountered this constraint before but it definitely makes sense in a situation when dealing with passengers.

This would indeed require to adjust `TWRoute::is_valid_addition_for_tw` to discard invalid break positions when adding jobs/breaks along a modified route segment. The tricky part is that I think some validity checks should happen directly in the body of `is_valid_addition_for_tw` while others should take place in `order_choice`, which is called [here from `is_valid_addition_for_tw`](https://github.com/VROOM-Project/vroom/blob/426263e387b4f0f8af0dce48ac9996fb80f8e934/src/structures/vroom/tw_route.cpp#L728-L729) and embeds the whole logic of job vs break insertion decision.
In particular, this would require to make sure that the adjustments to `order_choice` do not result in ruling out valid break insertions down the line.

jcoupey:
Feature-wise, we could see the concept of "only possible with empty vehicle" as a special case of "only possible under `X` load" with `X=[0...]`. The latter constraint would not be harder since both require checking against the current vehicle load. The more generic version could prove useful in other situations, something along the line of "wait until the truck is half-empty before doing the break".

fbaierl:
[...]

Issue #753, opened 2022-08-02
Force vehicle_step order 


fbaierl:
I think I may have misunderstood how exactly `vehicle_step`s are treated. My expecation was that if I give vehicle-steps to a vehicle, it will plan them in that order and use the remaining time-slots to plan the other given shipments.

As such I created an input like this:

- 1 vehicle (vehicle0)
- 6 vehicle steps
- 1 shipment (which clashes with my vehicle steps, so it should be 'unassigned')

```JSON
{
  "vehicles": [
    {
      "id": 0,
      "description": "vehicle0",
      "start": [
        12.304373066846503,
        51.62270653765847
      ],
      "end": [
        12.304373066846503,
        51.62270653765847
      ],
      "capacity": [7],
      "skills": [],
      "time_window": [
        1644188400,
        1644274740
      ],
      "breaks": [],
      "steps": [
        {
          "id": 0,
          "type": "job"
        },
        {
          "id": 1,
          "type": "job"
        },
        {
          "id": 2,
          "type": "job"
        },
        {
          "id": 3,
          "type": "job"
        },
        {
          "id": 4,
          "type": "job"
        },
        {
          "id": 5,
          "type": "job"
        }
      ]
    }
  ],
  "shipments": [
    {
      "amount": [3],
      "skills": [],
      "pickup": {
        "id": 7,
        "service": 120,
        "location": [
          12.175563,
          51.694164
        ],
        "time_windows": [
          [
            1644199440,
            1644199440
          ]
        ],
        "description": "pickup|1001"
      },
      "delivery": {
        "id": 7,
        "service": 120,
        "location": [
          12.219437,
          51.715622
        ],
        "time_windows": [
          [
            1644198918,
            1644198918
          ]
        ],
        "description": "delivery|1001"
      }
    }
  ],
  "jobs": [
    {
      "id": 0,
      "pickup": [0],
      "delivery": [0],
      "description": "",
      "location": [
        12.304373066846503,
        51.62270653765847
      ],
      "timeWindow": [
        [
          1644197801,
          1644197801
        ]
      ]
    },
    {
      "id": 1,
      "pickup": [0],
      "delivery": [0],
      "description": "",
      "location": [
        12.219437,
        51.715622
      ],
      "timeWindow": [
        [
          1644198798,
          1644198798
        ]
      ]
    },
    {
      "id": 2,
      "pickup": [2],
      "delivery": [0],
      "description": "pickup|5",
      "location": [
        12.219437,
[...]

PR #794, opened 2022-09-29
Add max load constraint for break validity

jcoupey:
## Issue

This PR aim at implementing #786.

## Tasks

 - [x] Add a `max_load` member for `Break` objects
 - [x] Parse `max_load` values in json input
 - [x] Assert `max_load` is OK when formatting routes
 - [x] Add `max_load` related violation in plan mode
 - [x] Pass delivery for insertion range as new arg in `is_valid_addition_for_tw`
 - [x] Maintain current load value while going through insertion range in `is_valid_addition_for_tw`
 - [x] Pass delivery for insertion range as new arg in `replace`
 - [x] Maintain current load value while going through insertion range in `replace`
 - [x] Add necessary checks in `TWRoute::order_choice`
 - [x] Add necessary checks in `TWRoute::is_valid_addition_for_tw`
 - [x] Maintain forward/backward minimum margin for breaks wrt `max_load` and current route load
 - [x] Use previous margins in `is_valid_addition_for_tw` validity checks to check for breaks outside the modified range
 - [x] Benchmark
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
There is a subtlety I didn't see at first in the adjustments required to update `is_valid_addition_for_tw`. The point of this function is to check validity while trying to replace an existing range of jobs with a new inserted range. We need to add a way to maintain the current route load during this process to check against the new break constraint.

The new load before insertion should be `1 - 2 + 3` with:

1. previous route load at that point;
2. total of job deliveries for replaced range (stuff loaded at startup that is not in the car any more after removing the old jobs);
3. total of job deliveries for insertion range (stuff that will have to be loaded at startup after inserting the new jobs).

Value 1 is already stored by the `RawRoute` object, while 2 can be computed **in constant time** using the `delivery_in_range` member function. Now in the `is_valid_addition_for_tw` context, the naive way to get 3 would be to go through the insertion range and sum the delivery values. But this would incur an expensive lookup (linear in the size of the insertion range) inside a hot spot.

The best way to avoid this I can think of is to provide this value as an additional argument to `is_valid_addition_for_tw`, since the call sites have better ways (read constant time) to get it. For example in local search operators, that value is already used for capacity checks in the `cvrp` version of the operator so it could be stored as a member for re-use in the `vrptw` context, or directly recomputed.

This will add a bit of boilerplate to the `is_valid_addition_for_tw` calls but I'm afraid there's no way around it.

jcoupey:
We'll also have to apply to `TWRoute::replace` the same changes as the ones already applied to `TWRoute::is_valid_addition_to_tw`, since both will have to pass the current load to `order_choice` in a consistent way. I updated the task list accordingly.

jcoupey:
There is yet another thing I under-estimated here: the changes already applied to `is_valid_addition_to_tw` and `replace` mean that we should not break the `max_load` constraint for a break taking place during the replaced/inserted range (we'll either place the break at a convenient place or decide that the move is invalid). This is necessary but not sufficient in general: replacing a range with another that has more job deliveries would impact the whole vehicle load from route start up to the beginning of the replaced range. Symmetrically, if the inserted range has more jobs pickups that the replaced range, load is increased for the whole end of the route. In short we need more checks for not breaking `max_load` for breaks *before* and *after* the insertion range.
In order to avoid looping though all breaks for those checks, we could maintain forward/backward min margins for breaks based on actual route load and `max_load` values.

This limitation only applies for instances with jobs, so normally the current state of this PR should be already working for shipment-only instances (ping @fbaierl if you want to give it a spin).

fbaierl:
Hello @jcoupey, not sure if I understand this right, but will this solution only work if there are no job elements inside the query at all? 

We actually use jobs (always with **empty pickup and delivery values**) to force the vehicles to maintain certain previously calculated routes (as written [here](https://github.com/VROOM-Project/vroom/issues/753#issuecomment-1250010349)). 

I have added an example to the original [issue](https://github.com/VROOM-Project/vroom/issues/786) with those empty jobs.

jcoupey:
> but will the current solution only work if there are absolutely no job elements inside the VROOM query at all?

I kind of made it sound like this but I should have said: "This limitation only applies for instances with jobs" *with non-zero pickup and deliveries*. Until I add the checks mentioned above, moving around jobs with non-zero pickup/delivery values can actually lead to invalid solutions (currently resulting in a crash on an `assert` in `format_route` after solving).

>We actually use jobs (always with empty pickup and delivery values)

In this case the current state of this PR *should* work using `max_load` keys for breaks.

jcoupey:
@fbaierl I just tried e536743 on the 3 instances you shared in #786. All instances result in breaks being taken with a non-empty load. Now adding a `"max_load": [0]` or `[0,0]` constraint to the break changes the solution toward a new one where breaks are taken with an empty load.

jcoupey:
I'm currently running some testing and I've spotted a few problematic situations along the way, such as P&D scenarios where we fail to insert shipments before or between the breaks, probably related to the heuristic decisions in `order_choice`. Not quite clear to me yet but I *think* this is due to a pre-existing bias, in which case this would already be affecting the current solving approach even before this PR. I'll investigate further and open a separate issue if necessary.

jcoupey:
I think the adjustments to `is_valid_addition_for_tw` and `order_choice` are now OK to handle `max_load` break values. Our new problem is that there are some places in the existing codebase where the checks should only be about timing since the `max_load` part does not make sense.

One example is when we test a single [pickup addition](https://github.com/VROOM-Project/vroom/blob/16c0e9c045d2d41611ddfebe946ff7424da61327/src/algorithms/local_search/insertion_search.h#L131) or [delivery addition](https://github.com/VROOM-Project/vroom/blob/16c0e9c045d2d41611ddfebe946ff7424da61327/src/algorithms/local_search/insertion_search.h#L108) in `insertion_search.h`. Of course inserting only a pickup or only a delivery does not make sense and never happens, but testing additions in isolations *for TW* is a great way to short-circuit a lot of loop iterations in case of time incompatibility. But now with the `max_load` checks added, any "single pickup" insertion will be seen as impossible before any break with a `max_load: [0]` since the load is never delivered, even though P -> D -> Break would not be a problem. The same applies the other way around for deliveries, resulting in obviously wrong solutions with a lot of unassigned jobs in situations with multiple breaks (despite vehicle idle time).

So what I tried in the previous commits is to add a flag as parameter to `is_valid_addition` and `order_choice` to switch on or off the `max_load` checks. I don't find the solution especially elegant and it makes both functions a bit more complex overall. But at least now we can pick the right behavior based on what is really required in a given context.

A benefit of this approach is that we could try storing at vehicle level whether the `max_load` checks are required at all, and override that flag dynamically. The idea is that we could then reduce the overhead of the changes in this PR for other "normal" instances without `max_load` constraints. This overhead is mostly related to passing amount information to the two functions and maintaining current route load while processing route portions for the validity checks.

PR #795, opened 2022-10-05
Chrono cleanup

kkarbowiak:
## Issue

Since `Timeout` optionally contains `std::chrono::milliseconds` the constructor calls are no longer needed.

## Tasks

 - [ ] ...
 - [ ] review


jcoupey:
Good point, thanks.

PR #796, opened 2022-10-05
Minor code cleanup

kkarbowiak:
## Issue

Just a bit of code cleanup.

## Tasks

 - [ ] ...
 - [ ] review

### DIFF
--- a/src/algorithms/validation/check.cpp
+++ b/src/algorithms/validation/check.cpp
@@ -7,8 +7,10 @@ All rights reserved (see LICENSE).
 
 */
 
+#include <algorithm>
 #include <mutex>
 #include <thread>
+#include <vector>
 #include <unordered_map>
 #include <unordered_set>
 
--- a/libvroom_examples/libvroom.cpp
+++ b/libvroom_examples/libvroom.cpp
@@ -98,7 +98,7 @@ void run_example_with_osrm() {
   // Create one-dimension capacity restrictions to model the situation
   // where one vehicle can handle 4 jobs with deliveries.
   problem_instance.set_amount_size(amount_dimension);
-  vroom::Amount vehicle_capacity(1);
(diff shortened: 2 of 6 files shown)


src/algorithms/validation/check.cpp
```diff
@@ -7,8 +7,10 @@ All rights reserved (see LICENSE).
 
 */
 
+#include <algorithm>
```
jcoupey, line 10: About those "missing" includes: my understanding is that we're kind of fine when another internal include does have the required ones. For example, most files end up including `typedefs.h` (directly or not) so they actually have `optional`, `string`, `vector` at hand. This assumption is probably a bit fuzzy though since we've had reports of compilation problems with such "missing" includes in the past.

Did you spot those based on compilation errors with some toolchain? Or is it just a common good practice to explicitly include everything? If so did you use a specific tool to spot the missing ones?

kkarbowiak: Well, one could argue that it is ok as long as the code compiles and it does :-) On the other hand if you remove an include from some header file (e.g. `typedefs.h`) due to a refactor, this could case compilation errors in some source files that depend on that include but do not include themselves. This can become tedious to fix.

I think it is a good practice to explicitly include everything, and there are tools to help with this, like [include-what-you-use](https://include-what-you-use.org/) (which by the way has a [rationale](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md)).

Anyway, the missing includes I added are result of eyeballing the codebase and are an ad-hoc effort. I did not use any tool.

jcoupey: Thanks for the pointers, makes total sense. I'd be in favor of running a more general cleanup campaign, which I ticketed in #797.

PR #807, opened 2022-10-21
Fix heuristic time going past timeout

### LINKED ISSUE #792: Timeout option has no effect
It seems that timeout option (the third parameter to `vroom::Input::solve` function or `--limit` command-line parameter to the stand-alone binary) has no effect.

I have a problem that takes about 45 minutes to solve using vroom. I tried to use the timeout option to see how much the results will differ if I want to have the results earlier. I tried passing the timeout (1000 and then 600 seconds), but even with it I still had to wait 45 minutes for the results.

I tried binary built using:
 * latest master ([b0fcb1b](https://github.com/VROOM-Project/vroom/commit/b0fcb1b0a3f726eb78e15e8d055f6f3b34c68ca0))
 * v1.12.0 release
 * v1.11.0 release (when the option was introduced)

I tried the three binaries on Manjaro.

Can someone please confirm (@jcoupey maybe)? I can give some more details if needed and am willing to try and investigate more.

jcoupey:
## Issue

This PR cherry-picks stuff from #755 that I tried back then but discarded at the time. The idea is to:

- eventually perform the solution filtering suggested in #750;
- fix #792.

## Tasks

 - [x] Split multithreading part in two: first heuristics then local search
 - [x] Spot and remove identical heuristic solutions based on solution indicators values
 - [x] Update `CHANGELOG.md`
 - [ ] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Use `std::chrono::milliseconds` for `Timeout` value (#728)
 - Use `struct` for storing `Coordinates` instead of an `std::array` (#730)
 - Refactor `SolutionIndicators` struct (#750)
+- Do not duplicate local search for identical heuristic solutions (#750)
 - Add message on invalid routing response (#764)
 - Consistent exception type on invalid profile (#771)
 - Pass zero amount directly instead of its size (#776)
@@ -23,6 +24,7 @@
 - Missing break validity check (#754)
 - Mark `JobAmount` and `JobTime` comparison operators as `const` (#724)
 - Update `ssl_send_and_receive` to throw RoutingExceptions (#770)
+- Timeout not observed with multiple long heuristics per thread (#792)
 
 ## [v1.12.0] - 2022-05-31
 
--- a/src/problems/cvrp/cvrp.cpp
+++ b/src/problems/cvrp/cvrp.cpp
@@ -7,7 +7,10 @@ All rights reserved (see LICENSE).
 
 */
 
+#include <iostream>
+
 #include <mutex>
+#include <set>
 #include <thread>
 
 #include "algorithms/heuristics/heuristics.h"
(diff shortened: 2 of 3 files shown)


kkarbowiak:
It looks like both `CVRP::solve()` and `VRPTW::solve()` functions have grown and became more (too) complex, but we can remedy this with a couple steps.

It seems both functions are identical, except for two details:
 1. `CVRP::solve()` starts with a part that falls back to simpler TSP problem.
 2. `CVRP::solve()` operates on `RawRoute` and `RawSolution` types, while `VRPTW::solve()` operates on `TWSolution` and `TWRoute` respectively.

Regarding 1. Could this check be extracted to `Input::get_problem()` function, so it returns the simpler problem instance?

Regarding 2. We could have a template function parametrised with `Route` and `Solution` types that would handle both cases. This would replace two functions with one. A further improvement could be gained by extracting handling threads to a separate class.

I can make some commits on top of this branch to show what I have in mind.

jcoupey:
You're right, the reason for having nearly similar duplicated implementation in the first place is that they were simple enough to not bother going through an additional abstraction layer. It would be a nice refactor to change that now.

>Regarding 1. Could this check be extracted to Input::get_problem() function, so it returns the simpler problem instance?

Definitely would make sense. We'd simply have to add a `TSP(const Input& input)` ctor.

Agree on using templates for the core solving code. Another thing to consider here is that we use 4 different parameter sets for all cvrp/vrptw X homogeneous/heterogeneous options.

Also happy to have the threads code live somewhere else if it makes the code more expressive and easy to read.

> I can make some commits on top of this branch to show what I have in mind.

Since this PR was originally a fix and we're talking about a more thorough refactor, I'd be in favor of merging as is now and moving the refactor part to another issue, what do you think?

kkarbowiak:
> Since this PR was originally a fix and we're talking about a more thorough refactor, I'd be in favor of merging as is now and moving the refactor part to another issue, what do you think?

Makes sense.

Issue #812, opened 2022-11-02
Duplicated code in `TSP::cost()` and `TSP::symmetrized_cost()` functions


kkarbowiak:
Both functions are exactly the same, but operate on different cost matrices. The common code can be extracted to a function parametrised with matrix to use.

jcoupey:
Good catch, always nice to deduplicate some code!

jcoupey:
Done in #813.

PR #813, opened 2022-11-02
Refactor TSP cost functions

kkarbowiak:
## Issue #812

Extracted common code to a separate function.

## Tasks

 - [x] review

### DIFF
--- a/src/problems/tsp/tsp.h
+++ b/src/problems/tsp/tsp.h
@@ -34,6 +34,8 @@ private:
   Matrix<Cost> _symmetrized_matrix;
   bool _round_trip;
 
+  Cost compute_cost(const std::list<Index>& tour, const Matrix<Cost>& matrix) const;
+
 public:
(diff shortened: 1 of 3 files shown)


src/problems/tsp/tsp.h
```diff
@@ -34,6 +34,8 @@ class TSP : public VRP {
   Matrix<Cost> _symmetrized_matrix;
   bool _round_trip;
 
+  Cost compute_cost(const std::list<Index>& tour, const Matrix<Cost>& matrix) const;
```
jcoupey, line 37: It looks like `compute_cost` does not even need to be a member function since I don't think it uses any other member from `TSP`. We could avoid exposing it altogether in the header and simply having it lie in the `.cpp` file as a non-member.

jcoupey (commented):
Thanks for the PR!

jcoupey (approved):

Issue #816, opened 2022-11-03
Update Ubuntu and compiler versions in CI builds


jcoupey:
Currently our CI builds run on `ubuntu-20.04` which comes with `g++-9` and `clang++-10`. It would be a good move to update to Ubuntu 22.04 for the image and use the new compilers versions available on that system. This also means updating the [code formatting script](https://github.com/VROOM-Project/vroom/blob/master/scripts/format.sh).

jcoupey:
Done in #817.

PR #817, opened 2022-11-03
Extend CI strategy to include g++ and clang++ installed on Ubuntu 22.04

kkarbowiak:
## Issue #816

I added Ubuntu 22.04 to the CI build matrix using both g++ and clang++, but without specifying their versions. According to [this table](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md) Ubuntu 22.04 image on GitHub includes clang++ 12, 13, and 14 and g++ 9, 10, and 11. Without specifying version the latest one is used (that is, clang++14 and g++11).

## Tasks

 - [x] review

### DIFF
--- a/.github/workflows/vroom.yml
+++ b/.github/workflows/vroom.yml
@@ -16,8 +16,24 @@ jobs:
   vroom:
     strategy:
       matrix:
-        cxx: ['g++-9', 'clang++-10']
-    runs-on: ubuntu-20.04
+        config:
+          - {
+              os: ubuntu-20.04,
+              cxx: 'g++-9'
+            }
+          - {
+              os: ubuntu-20.04,
+              cxx: 'clang++-10'
+            }
+          - {
+              os: ubuntu-22.04,
+              cxx: 'g++'
+            }
+          - {
+              os: ubuntu-22.04,
+              cxx: 'clang++'
+            }
+    runs-on: ${{ matrix.config.os }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


jcoupey:
Thanks for picking this up! I'd be in favor of simply updating the Ubuntu version to 22.04 (and explicitly pointing to the newest compiler versions) since:
- we don't commit to any extensive build checks platform-wise;
- we'll have to drop support/checks for the oldest compilers at some point anyway, so it may be now;
- doubling the number of CI builds sounds like a waste of time and resources for a small benefit;
- it would keep the workflow simpler.

What do you think?

kkarbowiak:
All of this makes sense to me. I left the old checks as I was not sure whether there was some commitment to support the particular OS/compiler version combination. Since there is none, we can simply use the latest versions. I can imagine this will closely match what people use as their development environment anyway.

jcoupey (approved):
The new builds passed seamlessly so we're fine. Can you just apply the same change to the `vroom_libosrm.yml` workflow? It should trigger a new `libosrm` build and cache both `g++-9` and `clang++-10` versions for further use.

jcoupey:
Hum, I should have anticipated this would fail since OSRM v5.26.0 has some outdated dependencies and does not compile any more on Ubuntu 22.04. Making the switch to v5.27.1 should solve this problem, and it's also a good move anyway.

kkarbowiak:
Yeah, I was wondering what the problem was.

jcoupey:
@kkarbowiak when you have time, can you push a commit changing the `env.osrm-tag` to v5.27.1 to see if it works?

Issue #818, opened 2022-11-03
Reduce code duplication in `CVRP` and `VRPTW`


kkarbowiak:
As discussed in #807, the common code in `CVRP::solve` and `VRPTW::solve` functions can be extracted to a separate template function.

jcoupey:
Landed with #819. Thanks @kkarbowiak!

PR #819, opened 2022-11-03
Refactor `CVRP::solve` and `VRPTW::solve`

kkarbowiak:
## Issue #818

Moved common code to a separate template function. I placed this function in `protected` section of the `VRP` base class, but perhaps it would be better to make it a non-member function placed in its own header file.

## Tasks

 - [ ] review


src/problems/vrp.h
```diff
@@ -10,17 +10,197 @@ All rights reserved (see LICENSE).
...
 namespace vroom {
 
-class Input;
-
 class VRP {
   // Abstract class describing a VRP (vehicle routing problem).
 protected:
   const Input& _input;
 
+  template <class TRoute, class TSolution, class TLocalSearch>
```
jcoupey, line 29: Pure nitpicking: in other template functions we use `Route` as template type to stand for either `RawRoute` or `TWRoute` so I'd be in favor of simply removing the `T` in front of template types here.

jcoupey (approved):
Thanks, this will really ease maintenance for this part of the code!

Having `solve` as a `VRP` member function makes total sense IMO as it factors in the generic approach used by both child classes.

src/problems/vrp.h
kkarbowiak: I added the `T` prefix (which I also dislike) just to avoid clash between the `Solution` template type and the `Solution` function return type, and for consistency applied it to `Route` and `LocalSearch` as well.

I can
- rename the `TSolution` template parameter to e.g. `CandidateSolution` and drop the `T` prefixes, or
- remove the `TSolution` template parameter and replace its use throughout the function with `std::vector<Route>`

What do you think?

jcoupey: Right, I did not think about the `Solution` collision. I like your second option since it removes a template type that can be trivially deduced from the other one. We can then also get rid of the `RawSolution` and `TWSolution` aliases in `cvrp.cpp` and `vrptw.cpp`. Another benefit is that it's safer to some degree as it prevents from using `solve` with inconsistent route and solution types.

PR #829, opened 2022-11-17
Refactor cost wrapper

### LINKED ISSUE #828: Refactor `CostWrapper`
We've introduced `CostWrapper` to be able to scale travel times on the fly based on vehicles `speed_factor` values. The current implementation is that we compute a scaling factor in the ctor:

https://github.com/VROOM-Project/vroom/blob/4b898ecc55d3005071e22b327dfe1d253c0d7ba1/src/structures/vroom/cost_wrapper.cpp#L17

Then when a duration value is requested, we scale the original matrix value (`DIVISOR` is 100 to keep a fair amount of precision since `speed_factor` is a double):

https://github.com/VROOM-Project/vroom/blob/4b898ecc55d3005071e22b327dfe1d253c0d7ba1/src/structures/vroom/cost_wrapper.h#L36-L39

The advantage of this approach is that even if `speed_factor` is a double, it keeps all our computations in the integer world (see #491). The current implementation have drawbacks though:
- we're still doing an integer division for each call to `duration` or `cost` (heavily used!);
- we still have rounding issues when getting back to seconds by dividing: depending on the `speed_factor` value, this can lead to different matrix durations being scaled to the same value.

I'd like to try to get rid of the division and simply use scaled values all over internally. This would amount to describing all durations as hundredth of second internally. We'd have to apply that change to all time-related values **and** round back to seconds upon solution formatting. This would add some code boilerplate but could improve precision while reducing computing time.

jcoupey:
## Issue

This PR is an attempt to fix #828.

## Tasks

 - [x] Get rid of division in `CostWrapper::cost` and `duration`
 - [x] Scale internal times received in input
 - [x] Get back to original values after solving in all `vroom::Solution`-related objects
 - [x] Check for non-regressions
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
Scaling input durations internally means we have to also scale values in time windows. The problem is that when using regular timestamps like 1669022846, scaling by a factor of 100 will quickly bring us over the limit for a `uint32_t`. My first attempt in 6d38cc6 has been to switch `Cost` and `Duration` types to `uint64_t`. It works but is not ideal since it means that we'll store matrices that are basically twice the size in RAM for no benefit (the travel time values themselves fit in a `uint32_t`).

To avoid the memory overhead my plan is to refactor types further so that:
1. `Cost` and `Duration` are only user-facing types used in the json and C++ API, they stay `uint32_t`.
2. Everything cost or duration-related internally is computed as a signed int (`uint64_t` gives us plenty of room).
3. Everything cost or duration-related internally is *stored* as a signed int.

That last point is the one requiring the most changes, but as soon as everything happens in the signed world with enough room, we should be safe from the kind or problem spotted in #831, while not increasing our memory footprint.

jcoupey:
> nothing changed from the API usage (both json and C++)

Well, almost: as the last CI build failure shows, users of the C++ API that provide custom matrices will have to switch from `vroom::Matrix<vroom::Duration>` to `vroom::Matrix<vroom::UserDuration>` (same underlying type).

jcoupey:
This adds a lot more boilerplate than I first though it would, especially in order to make sure reported results are consistent in output after scaling back to `UserCost` and `UserDuration` values. But I think it's worth the trouble since:

- it will open up to further scaling options (required in #788);
- getting rid of the integer division for values scaling upon each call to `cost` or `duration` does seem to improve computing time.

On that last point, here are some values for average instance computing time per benchmark class, using `-x 5` on the usual VRPTW benchmarks.

Instances | `master` | PR at f47a0368 | Delta (%)
-- | -- | -- | --
Solomon | 459 | 436 | -5.0
Homberger 200 | 1889 | 1794 | -5.0
Homberger 400 | 9215 | 8706 | -5.5
Homberger 600 | 22562 | 21371 | -5.3
Homberger 800 | 44047 | 42003 | -4.6
Homberger 1000 | 76170 | 72887 | -4.3

PR #901, opened 2023-03-11
pyvroom: expose speed_factor and per_hour

jonathf:
I'm updating pyvroom to get feature pairity with version 1.13. To get there I need to expose the speed_factor attribute.

I'm also exposing per_hour, thought I do have it indirectly in other context, I think it is much cleaner if it is exposed it directly aswell.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changed
 
+- Exposed internal variables to get feature parity for pyvroom (#901)
 - Update GitHub Actions (#857)
 - Improve error messages (#848)
 - Add running `apt-get update` in CI jobs (#863)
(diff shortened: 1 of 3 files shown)


jcoupey:
No problem on the principle of exposing those two. My only concerns are:

- I'm reluctant to tag a patch release only for a "convenience" change for downstream use;
- if we merge this into the `release/1.13` branch, we'd have to also apply it to `master`.

What do you think if we do it the other way around: we merge those changes into current `master` first to have them in main line, then I simply cherry-pick what's required to the `release/1.13` branch. Maybe we don't even have to tag a release if you aim at the tip of that branch from downstream?

jonathf:
Thats fine.

I've rebased the branch and repointed the PR towards master now.

I also recompiled pyvroom against the current master, and it works the same as the release. I have no problem pointing pyvroom to the current tip.

jcoupey:
I just cherry-picked the changes from this branch to `release/1.13` in e2653a46.

PR #911, opened 2023-04-03
Setup clang-tidy

### LINKED ISSUE #789: Using clang-tidy
We could probably benefit from running [`clang-tidy`](https://clang.llvm.org/extra/clang-tidy/) on the codebase. Not sure whether we want to just give it a go and fix some stuff, or enable that in CI checks. Probably worth a try anyway.

jcoupey:
## Issue

Fixes #789 

## Tasks

 - [x] Decide on which checks to use (https://clang.llvm.org/extra/clang-tidy/checks/list.html)
 - [x] Apply relevant changes to the codebase
 - [x] Provide a tidying script in some way
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
Trying to set this up carefully is a painful journey... But it did allow for some refactoring already and hopefully it will be useful in the long run.

Since the whole checking process is quite long, we don't really want to run it very often, not in CI anyway. I just added a simple makefile target to run the checks. It's probably not really portable since system include paths are hardcoded, but it will do for now.

There are a few remaining checks on which I have doubts, questions, or that would require much bigger overhaul. In order to keep this first setup simple but not loose the relevant info, I'll open a dedicated ticket to list concerns and ideas for improvements.

Issue #931, opened 2023-06-20
Update polylineencoder


jcoupey:
Required for #851.

jcoupey:
Landed with #932.

PR #932, opened 2023-06-21
Switch to C++20

kkarbowiak:
## Issue #851 #931

Another take at switching to C++20.

## Tasks

 - [ ] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Update formatting script to use `clang-format` 14 (#894)
 - Setup a `clang-tidy` workflow (#789)
 - Submodule and update Rapidjson (#929)
+- Update polylineencoder to v2.0.1 (#931)
 - Switch to C++20 (#851)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Update formatting script to use `clang-format` 14 (#894)
 - Setup a `clang-tidy` workflow (#789)
 - Submodule and update Rapidjson (#929)
+- Switch to C++20 (#851)
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


kkarbowiak:
> Thanks for the new PR! Looks like everything passed as expected on the CI side. Oddly, I get an error when trying to build locally using clang:
> 
> ```
> $ export CXX=clang++-14
> $ make -j 1
> clang++-14 -MMD -MP -I. -std=c++20 -Wextra -Wpedantic -Wall -O3 -DASIO_STANDALONE -DUSE_ROUTING=true -D USE_LIBGLPK=true -c main.cpp -o main.o
> main.cpp:10:10: fatal error: 'fstream' file not found
> #include <fstream>
>          ^~~~~~~~~
> 1 error generated.
> make: *** [makefile:76: main.o] Error 1
> ```
> 
> The weird thing is I have the same setup as for the CI builds (Ubuntu 22.04) and the issued command [is identical](https://github.com/VROOM-Project/vroom/actions/runs/5333719841/jobs/9664561398?pr=932#step:4:7). The only hint is that all errors seem to relate to includes from the standard library.
> 
> EDIT: I actually get the same errors on current `master` with clang so the problem is not specific to this PR.

Strange!

I am using Ubuntu 22.04 on WSL2 and cannot reproduce this problem. The build is fine either specifying exact version (CXX set to clang++-14) or just the compiler name (CXX set to clang++). Both invocations use the same binary, anyway.

jcoupey:
>I am using Ubuntu 22.04 on WSL2 and cannot reproduce this problem

Yeah, it works on CI and I also built successfully using `clang++-14` on another box using Ubuntu 22.04. Maybe I just messed up something on my laptop. And again the same happens on `master` so this is definitely not C++20-related, so not a blocker for merging anyways.

PR #938, opened 2023-06-22
Apply clang-tidy changes

kkarbowiak:
## Issue

These are new changes after switching to C++20.

## Tasks

 - [ ] review

### DIFF
--- a/src/problems/tsp/heuristics/local_search.cpp
+++ b/src/problems/tsp/heuristics/local_search.cpp
@@ -10,6 +10,7 @@ All rights reserved (see LICENSE).
 #include <algorithm>
 #include <iterator>
 #include <numeric>
+#include <ranges>
 #include <thread>
 #include <unordered_map>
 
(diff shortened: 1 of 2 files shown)


jcoupey:
Interesting. Can you share how you actually run the checks?

kkarbowiak:
> Interesting.

I posted this PR draft just to share these strange results I also found on my box. Curiously, clang fails to build changes suggested by clang-tidy, while gcc compiles without a fuss. Go figure :-)

> Can you share how you actually run the checks?

I executed `make tidy` followed by `make -j` and got the same failure using clang++.

kkarbowiak:
I'm closing this PR as it cannot be merged.

jcoupey:
Actually I'd like to be able to switch to using ranges at some point for most of the situations where we currently have iterator-based functions. But the suggested changes here are not the first that would come to mind. ;-)

PR #967, opened 2023-08-08
allow a host to have a path

### LINKED ISSUE #966: Path in URL won't work for HTTP router interface
Similar to #947 , but this time it's an actual issue ;)

E.g. OSRM with on https://routing.openstreetmap.de/routed-car can't be used as host input, the same for any other backend that has a URL path other than the host's root. I'll PR shortly:)

nilsnolde:
## Issue

fixes #966 

Now this can be done (note, we're robust to (non-)existing trailing slash in the path):
```
vroom -i docs/example_3.json -r osrm -a car:routing.openstreetmap.de/routed-car/ -p car:443 -c
```

## Tasks

 - [x] review

### DIFF
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -27,6 +27,7 @@ if type clang-format-14 2> /dev/null ; then
 elif type clang-format 2> /dev/null ; then
     # Clang format found, but need to check version
     CLANG_FORMAT=clang-format
+    V=$(clang-format --version)
     if [[ $V != *14.0* ]] ; then
         echo "clang-format is not 14.0 (returned ${V})"
         exit 1
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -49,7 +49,7 @@ int main(int argc, char** argv) {
     .set_tab_expansion()
     .add_options("Solving")
     ("a,host",
-     "host for the routing profile",
+     "host for the routing profile, optionally with a URL path, e.g 'routing.openstreetmap.de/routed-car'",
      cxxopts::value<std::vector<std::string>>(host_args)->default_value({vroom::DEFAULT_PROFILE + ":0.0.0.0"}))
     ("c,choose-eta",
      "choose ETA for custom routes and report violations",
--- a/src/structures/cl_args.cpp
+++ b/src/structures/cl_args.cpp
@@ -17,6 +17,7 @@ void update_host(Servers& servers, const std::string& value) {
   // Determine profile and host from a "car:0.0.0.0"-like value.
   std::string profile = DEFAULT_PROFILE;
   std::string host;
+  std::string path = "";
 
   auto index = value.find(':');
   if (index == std::string::npos) {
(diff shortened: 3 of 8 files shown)


scripts/format.sh
```diff
@@ -27,6 +27,7 @@ if type clang-format-14 2> /dev/null ; then
 elif type clang-format 2> /dev/null ; then
     # Clang format found, but need to check version
     CLANG_FORMAT=clang-format
+    V=$(clang-format --version)
```
nilsnolde, line 30: this was a copy/paste fail. I have `15.0.7`. would you mind running format.sh?

src/main.cpp
```diff
@@ -49,7 +49,7 @@ int main(int argc, char** argv) {
     .set_tab_expansion()
     .add_options("Solving")
     ("a,host",
-     "host for the routing profile",
+     "host for the routing profile, optionally with a URL path, e.g 'routing.openstreetmap.de/routed-car'",
```
nilsnolde, line 52: technically it's not a "host" anymore but IMO better than having a new argument

scripts/format.sh
jcoupey: Went unnoticed on my side since formatting still works, despite this check variable missing. Totally weird.

src/structures/cl_args.cpp
```diff
@@ -26,13 +27,27 @@ void update_host(Servers& servers, const std::string& value) {
     host = value.substr(index + 1);
   }
 
+  // remove any trailing slash
+  if (host.find('/') == host.length() - 1) {
+    host.pop_back();
+  }
```
jcoupey, line 33: This would fail to remove the trailing slash with multiple occurrences, say for a host like `routing.openstreetmap.de/the/osrm/slug/routed-car/`. We should probably rather simply check the value of `host.back()` (if `host` is not empty).

src/structures/cl_args.cpp
```diff
@@ -26,13 +27,27 @@ void update_host(Servers& servers, const std::string& value) {
...
+
+  // pull out a path if any and append a trailing slash for query building
+  index = host.find('/');
+  if (index != std::string::npos) {
+    path = host.substr(index + 1) + "/";
```
jcoupey, line 38: Passing an empty host used to be OK during update and generate a "clean" error upon connection error. On the other hand, this will break:
```
$ vroom -a "" ...
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)
```

Wrapping lines 30-40 in a `if (!host.empty())` clause should do the trick?

jcoupey (commented):
Thanks for the PR, mentioned a couple edge cases in the comments.

jcoupey:
This part of the code is not touched really often, but I just merged a couple PR that introduced conflicts here. Sorry for the inconvenience...

src/structures/cl_args.cpp
nilsnolde: of course, my bad

src/structures/cl_args.cpp
nilsnolde: jep, fixed

nilsnolde:
Thanks for the review @jcoupey , comments are addressed and master merged.

src/structures/cl_args.cpp
```diff
@@ -26,13 +27,29 @@ void update_host(Servers& servers, const std::string& value) {
...
+    // remove any trailing slash
+    if (host.back() == '/') {
+      host.pop_back();
+    }
+
+    // pull out a path if any and append a trailing slash for query building
+    index = host.find('/');
+    if (index != std::string::npos) {
+      path = host.substr(index + 1) + "/";
+      host = host.substr(0, index);
```
jcoupey, line 40: Nitpicking: this is making a copy while we could simply `host.erase(index)`.

jcoupey (approved):
Thanks for the changes. One remaining improvement came to mind, I'll simply do this myself before formatting and merging.

src/structures/cl_args.cpp
nilsnolde: good point! done. haha but too late. feel free to change it before next release

PR #968, opened 2023-08-12
Improve `vrptw::PDShift`

### LINKED ISSUE #852: Unnecessary range copy in `vrptw::PDShift`
We're making a copy of the "remaining" source range in the `vrptw::PDShift` operator ctor:

https://github.com/VROOM-Project/vroom/blob/0aeb2ccfa5b4377a053e2934f122f8a3b3ef1d0b/src/problems/vrptw/operators/pd_shift.cpp#L35-L36

This means a lot of overhead for a downstream use where we only use iterators in validity checks. The tricky part is that the copy is indeed necessary in`apply` since we can't replace part of a route while iterating over a subset of the same route. But we could make it so that the copy only happens in `apply` which is only called a marginal number of times across all instances of the operator.

jcoupey:
## Issue

Fixes #852

## Tasks

 - [x] Only generate `source_without_pd` upon applying `PDShift`
 - [x] Remove unnecessary loops that can be replaced by `delivery_in_range`
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - Add job id to error message for unreachable step (#946)
 - Reduce `compute_best_route_split_choice` complexity (#962)
 - `Eval::operator<` sorts on cost, then duration (#914)
+- Improved `vrptw::PDShift` implementation (#852)
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


jcoupey:
The (limited) tests I ran on the smallest Li&Lim instances do not show a significant computing time gain here.

PR #972, opened 2023-08-21
Expose geometry

jonathf:
## Issue

There has been a request for accessing the geometry attribute:
https://github.com/VROOM-Project/pyvroom/issues/95

The attribute is unaccessible on the vroom side, so this patch exposes it so it can be read from pyvroom.

## Tasks

 - [ ] review

### DIFF
--- a/src/structures/vroom/input/input.cpp
+++ b/src/structures/vroom/input/input.cpp
@@ -44,6 +44,10 @@ void Input::set_geometry(bool geometry) {
   _geometry = geometry;
 }
 
+bool Input::get_geometry() {
+  return _geometry;
+}
+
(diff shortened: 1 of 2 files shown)


nilsnolde:
Oops, sorry, I just realized pyvroom is already doing everything properly.. I don't know how I didn't see that last night, but there's already Input::set_geometry in pyvroom which is exactly what it needs. It's not on Solution but on Input class..

I think a few things will have to be amended (instead of exposing `_geometry`):
- the Input::set_geometry of the bindings is currently a no-op, it needs to take a boolean to set the `_geometry`
- the `_routes_numpy` conversion needs to add the geometry strings if present

Probably a bit more than that. I can look into it soon too I'd update when working on it and let me know if you're first :)

Issue #971, opened 2023-08-17
Using priorities in route seeding


jcoupey:
Our main optimization objective is priority score (i.e. sum of priorities for assigned tasks), then number of assigned tasks, then the "usual" cost.

Currently the heuristics make no use of priority values whatsoever, so initial solutions can be arbitrarily bad in term of priority score (the heuristics were actually implemented even before `priority` was a thing for tasks). Then the local search tries to improve the priority score: priority is the foremost criterion for task insertion in `try_job_additions` and the `UnassignedExchange` operator will gladly switch an assigned task with another unassigned task having higher priority (and maybe higher cost).

In a situation with several priority levels in input and where priority is key, it could be beneficial that heuristics have incentives to favor higher-priority tasks. I don't think it would make sense to put that as the foremost criterion for all insertions since then we would on the other hand generate high-priority initial solutions that would be arbitrarily bad in term of cost. It may be even harder to fix by the local search this way around.
But the route seeding process in the heuristics would be a good place to make sure that we at least start any new route (except with `INIT::NONE`) with a high-level task.

jcoupey:
> Maximise the sum of the priorities    Maximise the number of job assigned    Minimise the cost

Yes, those are the optimization criteria, in that order.

> What is meant here by "usual" cost? Mileage ?

Outside the priority score and  number of assigned tasks, the cost we use as optimization objective is based on vehicles fixed costs (if any) and travel costs.

> Is that the only optimisation criteria ?

We should soon be able to ponderate distances with travel times in the overall notion of "cost", see https://github.com/VROOM-Project/vroom/issues/908#issuecomment-1484676727 on this.

jcoupey:
The idea from #988 implemented in #1021 is actually somewhat simpler and more efficient, so closing here.

Issue #988, opened 2023-09-13
Wider priority-targeted local search move


jcoupey:
This issue is related to #971, which proposes to improve priority scores relying on initial heuristic solution changes. A dual approach would be to still let the local search do the priority heavy lifting after overcoming the limitations of the current move options we have.

Currently adding higher-priority jobs happen in `try_job_additions` whenever we (re)populate modified routes, or via the `UnassignedExchange` operator that allows to replace a job with another higher-priority job, even if at a more expensive cost. This is somehow limited and replacing just one job is clearly not enough in many situations.

What we could do instead is try replacing a whole portion of route with an unassigned job so that priority increases. The more jobs we remove, the more chance we have that the switch will be valid constraint-wise. To cut down the complexity of choosing which jobs to remove, we could limit to series of consecutive jobs at the beginning or end of a route. We could then easily decide where to "cut" the route based on any priority value by storing cumulative priority values in routes.

jcoupey:
Landed with #1021.

PR #989, opened 2023-09-14
Experiment with priority initialization

jcoupey:
## Issue

This PR tracks experimental work on #971.

## Tasks

 - [x] Try various ways to account for priority in heuristics
 - [x] Benchmark
 - [ ] Update `CHANGELOG.md`
 - [ ] review


jcoupey:
I've been trying various options here (e.g. only account for priority in route seeding or for all insertions, use a dedicated `PRIORITY` seeding policy) and the results are not as good as one might have expected.

On a sample of priority-bound instances, sure the priority scores are better with this PR than with `master` **if one looks at initial solutions after the heuristic**. But after the local search phase, the difference is not significant, and even sometimes (actually quite often) not accounting for priority in heuristics lead to solutions that are actually better on all fronts: priority score, number of assigned jobs and overall cost!

My gut feeling here is that when starting from an initial solution that is poor in term of priority but good in term of cost, it's easier for our local search to improve on priority since there is "more room" for changes. On the other hand, if we have a solution that is good in term of priority in the first place, then it's usually much more expensive and more constrained, so the local search can't get away from the nearest cost local minima.

We should probably re-evaluate the interest of this PR at some point in the light of #988.

jcoupey:
Closing as this is superseded by #1021.

PR #990, opened 2023-09-14
SWAP* enhancement

### LINKED ISSUE #987: Reduce SWAP* computational burden
We've introduced the SWAP* local search operator back in #507, an efficient extension of our previous `Exchange` more naive operator.

I've been doing some low-level profiling recently and found out that the time spent on SWAP* is unexpectedly high in many situations. After giving this more thoughts, I think we could move a lot of the computing to store additional stuff in `SolutionState` and enjoy much more data re-use upon operator evaluation.

To be a bit more specific, whenever we evaluate SWAP* for routes `source` and `target`, there is a (quadratic) pre-processing step applied to find top 3 insertions of all jobs from `source` into `target` and vice-versa top 3 insertions of all jobs from `targert` into `source`. Now if a different operator only touches `target`, then for sure the first part has to be re-computed, but the "jobs from `target` into `source`" part is still valid. The flaw is that we later recompute that across all source routes after modifying `target`, while we should only update the "other jobs into `target`" part.

jcoupey:
## Issue

Fixes #987 

## Tasks

 - [x] Store `find_top_3_insertions` results in `SolutionState`
 - [x] Test using various gain upper bounds for early aborts
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Reserve `vector` capacity whenever possible (#915)
 - Refactor heuristics to be able to operate on a subset of jobs and vehicles (#837)
 - Account for vehicle/job compatibility in heuristic regrets values (#982)
+- Slightly reduce computing times for SWAP* operator (#987)
 
 ### Fixed
 
--- a/src/algorithms/local_search/local_search.cpp
+++ b/src/algorithms/local_search/local_search.cpp
@@ -179,6 +179,8 @@ void LocalSearch<Route,
     }
   }
 
+  std::unordered_set<Index> modified_vehicles;
+
   do {
     Priority best_priority = 0;
     RouteInsertion best_insertion(_input.get_amount_size());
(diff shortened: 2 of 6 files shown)


jcoupey:
Results so far are not overwhelming but still interesting. There are small differences in the solutions since baf8109 cuts some branches assuming the triangular inequality holds on costs. Computing times reported below with a dev build.

## Computing times (ms) on the usual CVRP benchmarks across `x` values:

`x` | `master` | baf81096 | Delta (%)
-- | -- | -- | --
0 | 349 | 324 | -7.2
1 | 1267 | 1128 | -11.0
2 | 2879 | 2672 | -7.2
3 | 5082 | 4830 | -5.0
4 | 9803 | 9305 | -5.1
5 | 15587 | 14916 | -4.3

## Computing times (ms) on the usual VRPTW benchmarks using `x=4`:

  | `master` | baf81096 | Delta (%)
-- | -- | -- | --
Solomon | 438 | 428 | -2.3
Homberger 200 | 1888 | 1782 | -5.6
Homberger 400 | 8804 | 8465 | -3.9
Homberger 600 | 20964 | 20315 | -3.1
Homberger 800 | 40057 | 39055 | -2.5
Homberger 1000 | 67054 | 66555 | -0.7

Issue #957, opened 2023-07-26
Simplify routing wrapper `add_route_info`


jcoupey:
Currently every routing wrapper implements the virtual `Wrapper::add_route_info` function to populate the route geometry and the `routes[].steps[].distance` values in output. This means the same logic is duplicated in various files to populate the distances, with small subtle differences depending on the underlying routing engine.

Once we have #956, we should be able to simply populate `Step::distance` values during the solution formatting process just as for `Step::duration`. Then we could turn the various `*::add_route_info` to much simpler `*::add_geometry` functions.

PR #991, opened 2023-09-18
Add `max_distance` constraint

### LINKED ISSUE #354: Maximum distance for vehicle
Will be great  to have maximum distance for the route in vehicle config . To disallow vehicle trips more then this distance.

jcoupey:
## Issue

Fixes #354.

## Tasks

 - [x] Parse `max_distance` value in input
 - [x] Store `max_distance` at vehicle level
 - [x] Adjust `Eval` and `CostWrapper` to hold distances
 - [x] Replicate `max_travel_time` logic in heuristics and local search to account for max distances constraints
 - [x] Adjust vehicle ordering in heuristics to account for `max_*` constraints
 - [x] Update `docs/API.md`
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/src/structures/vroom/input/input.cpp
+++ b/src/structures/vroom/input/input.cpp
@@ -862,6 +862,50 @@ void Input::set_vehicle_steps_ranks() {
   }
 }
 
+void Input::init_missing_matrices(const std::string& profile) {
+  // Even with custom matrices, we still need routing after
+  // optimization if geometry is requested.
+  bool create_routing_wrapper = _geometry;
+
+  if (const auto durations_m = _durations_matrices.find(profile);
+      durations_m == _durations_matrices.end()) {
+    // No custom durations matrix.
+
+    if (_distances_matrices.contains(profile)) {
+      // We don't accept distances matrices without durations
+      // matrices.
+      throw InputException(
+        "Custom matrix provided for distances but not for durations for " +
+        profile + " profile.");
+    }
+
+    // No durations/distances matrices have been manually set,
+    // create empty ones to allow for concurrent modification later
+    // on.
+    create_routing_wrapper = true;
+    _durations_matrices.try_emplace(profile);
+    _distances_matrices.try_emplace(profile);
+  } else {
+    // Custom durations matrix defined.
+    if (!_distances_matrices.contains(profile)) {
+      // No custom distances.
+      if (_geometry) {
+        // Get distances from routing engine later on since routing
+        // is explicitly requested.
+        _distances_matrices.try_emplace(profile);
+      } else {
+        // Routing-less optimization with no distances involved,
+        // fill internal distances matrix with zeros.
+        _distances_matrices.try_emplace(profile, durations_m->second.size(), 0);
+      }
+    }
+  }
+
+  if (create_routing_wrapper) {
+    add_routing_wrapper(profile);
+  }
+}
+
 void Input::set_matrices(unsigned nb_thread) {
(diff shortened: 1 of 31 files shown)


src/structures/vroom/input/input.cpp
```diff
@@ -602,9 +602,15 @@ void Input::set_vehicles_TSP_flag() {
...
-    vehicle.cost_wrapper.set_durations_matrix(&(d_m->second));
+    auto duration_m = _durations_matrices.find(vehicle.profile);
+    assert(duration_m != _durations_matrices.end());
+    vehicle.cost_wrapper.set_durations_matrix(&(duration_m->second));
+
+    // TODO this will probably break with user-provided durations but
+    // not distances.
+    auto distance_m = _distances_matrices.find(vehicle.profile);
+    assert(distance_m != _distances_matrices.end());
+    vehicle.cost_wrapper.set_distances_matrix(&(distance_m->second));
```
jcoupey, line 613: This indeed causes the CI build to break on the usual check with our `example_2.json` file because no distances matrix is provided along with the durations matrix. To shadow this problem, we could simply update example 2 with the distances matrix (it makes sense to do so anyway).

But we should decide on how to retro-compatibly handle the situation of custom matrices with durations (maybe costs) but no distances. I think the best default way of doing it would be to create a distances matrices filled with zeros. This way it would not impose any change on costs and constraints.

jcoupey:
The hard part about custom matrices in input is that it is not always clear what to do when durations are provided but not distances. It could be either that users expect to:

1. get distances from the routing engine;
2. not account for distances in the optimization as was the case previously.

Enforcing 1. would be a breaking change for "self-contained" instances (e.g. all usual benchmarks) with custom durations only that would not run any more without a routing engine. Enforcing 2. would keep the desired behavior for self-contained instances, but on the other hand it should be a valid use-case to only provide custom durations and expect distances to be fetched from the routing engine, then used in optimization.

So I went with deciding on 1. or 2. dynamically based on whether the `-g` flag is provided. If it is, it means that routing is expected anyway to fill in distances at route level, so we go with 1. If not, we assume that distances are not meaningful for optimization and **we set a distances matrix filled with zeros**.

jcoupey:
Follow-up note on custom matrices: we have no commitment so far to support custom distances so we can actually request that they're only provided if durations are also custom. This solves the dual version of the above question: providing only custom distances without custom durations is an input error.

jcoupey:
One thing I overlooked is that adding a `VIOLATION::MAX_DISTANCE` in plan mode is not trivial because it requires to account for accumulated distances throughout the route, which we only do later on currently. Actually the scope of #957 is precisely to move that computation during route building now that we have distance matrices at hand.

I think the simplest way to move forward is to first merge support for `max_distance` in solving mode (this PR), then handle the `MAX_DISTANCE` violation in a separate ticket as a follow-up of #957.

jcoupey:
With this PR, distance range for vehicle is based on `max_travel_time` keys and values in the distance matrix. Now currently (until we have #957 implemented), the actual distances reported in output in `routes` and `routes[].steps` are computed based on the subsequent `route` requests to the routing engine. Since there is some rounding involved and the rounding is done on summed values instead of each distance pair, this can result in discrepancies.

For example with this PR as is, it is possible to get a route distance in output slightly over the `max_travel_time` value. Not a bug, just that the limitation is done on the distance matrix values, not on the distances from the `route` request showing up in output. This kind of mismatch should vanish with #957.

Issue #986, opened 2023-09-08
Dig into sonarcloud code smells reports


jcoupey:
Now that the workflow is in place after #984 and the most severe reports are fixed, there are different follow-up tasks here:

1. identifying the rules we really want to comply to;
2. find a way to silence the rules we do not want to take into account;
3. update the codebase to gradually remove the remaining reports.

jcoupey:
So I did a lot of refactoring based on the code smell reports. There are probably remaining reports that could be easily fixed and I'll probably do that along the way with other PR. For now the most important remaining reports category are:

1. Control flow statements nesting
2. Use `std::format` instead of manual concatenations
3. Cognitive complexity of functions

I don't want to address 1. and 3. right now because they would require deeper changes. Maybe those are the right places to use the improve-as-you-code approach targeted with the "new code" score.

I'd like to handle 2. but sadly support for `format` is not available in gcc 11 which is our current reference version on CI.

jcoupey:
Closing here as I don't think we want to silence any reports for now. We can just live with them and try to fix them as we go.

PR #992, opened 2023-09-18
Address some sonar reports

jcoupey:
## Issue

This PR is a first take at #986, trying to get the low-hanging fruits with quick transparent changes that address a lot of reports.

## Tasks

 - [x] Fix most "consistency" reports
 - [x] Dig a subset of "Intentionality" reports
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
So one limitation here is that when analyzing the diff, sonar tells us it's crappy since the changes touch a lot of places with code smell reports without fixing them, so it kind of "inherits" the blame. As a result the PR quality gate is not met from a  "New Code" perspective. In theory this should prevent merging, but here I've addressed a lot of the code smells for "Overall Code" so merging would greatly reduce the code smells overall.

I guess I'm simply not adhering to the "change as you code" workflow here so I'll just merge and see how the reports go in the sonarcloud UI.

PR #1003, opened 2023-09-28
Replace g++-11 with g++-12 in CI

### LINKED ISSUE #1002: Update to gcc 12 in CI
We're experiencing a lot of failed CI builds with gcc lately. What happens is that the GA jobs are just stopped with `Error: Process completed with exit code 143.`, indicating the job is simply killed for some reason. The running times widely vary so this is not about some timeout. I stumbled on https://github.com/actions/runner-images/issues/6680 which is totally familiar.

Probably the recent changes we introduced (switching to C++20 and using more features from 20 following the sonarcloud reports) somehow increased the load on gcc, causing the builds to get killed.

Maybe switching to gcc 12 could help here. Also it would be an easy switch as version 12 is packaged for the Ubuntu images we're using on CI. Plus it's a good idea to update at some point anyway.

jcoupey:
## Issue

Fixes #1002 

## Tasks

 - [x] Update workflows
 - [ ] ~~Add install step for `g++-12`?~~
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/.github/workflows/vroom.yml
+++ b/.github/workflows/vroom.yml
@@ -16,7 +16,7 @@ jobs:
   vroom:
     strategy:
       matrix:
-        cxx: ['g++-11', 'clang++-14']
+        cxx: ['g++-12', 'clang++-14']
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout repository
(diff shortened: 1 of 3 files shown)


jcoupey:
Looks like this did the trick to fix the failing builds problems. We now get [cryptic warnings](https://github.com/VROOM-Project/pyvroom/issues/81#issuecomment-1621357809) from g++-12 but a bit of searching seems to indicate this is a bug in the compiler so we can probably live with it for now.

PR #1008, opened 2023-10-05
Work on reducing operators scope

### LINKED ISSUE #509: Reduce local search operators scope
The current way local search operators look up for neighboring solutions is a bit naive in that we exhaustively check all options for routes / pairs of routes. Obviously relocating a job in another route at a rank where tasks  are very far away does not make sense. In that case we stop early without any validity check because the gain itself is not interesting. Yet probably a lot of time is actually spent evaluating gain for silly moves.

I recall toying with more advanced filtering in the early days of the implementation, e.g. only trying a relocate move to the nearest routes or ranks in routes. From what I recall, it did have a detering impact on quality (some interesting moves missed in the process) and the computing gain was not that relevant because the overall approach was "fast enough" as a first implementation.

Now I think would be a good time to evaluate the speedups we could get from this idea. The question is whether we'll be able to easily find the sweet spot between unnecessary evaluation and missed moves. Either we can come up with some threshold where we'll only miss a marginal number of moves, either we can maybe make this configurable to some extent.

jcoupey:
## Issue

Fixes #509 

## Tasks

 - [x] Simple filtering rules at route pair level for `*Exchange`
 - [x] Simple filtering rules at route pair level for `SwapStart`
 - [x] Simple filtering rules at route pair level for `(Reverse)TwoOpt`
 - [x] Update `CHANGELOG.md`
 - [x] review


jcoupey:
After a routine check on HVRP instances, I noticed the solution quality were highly degraded with this PR. This makes sense because the pruning argument based on geometric consideration somehow hold only if we assume homogeneous costs between routes.

In order to keep the benefit of the improvements here but avoid missing too many interesting moves, we should fine-tune and only prune route pair moves if the associated vehicles are "similar".

jcoupey:
Turns out most of the HVRP instances (the DLP dataset) do not provide point locations... As a result the bbox checks where discarding **all** moves.

Nevertheless only pruning for homogeneous vehicle pairs is still a good idea since otherwise we're loosing too much quality.

Issue #623, opened 2021-12-08
Ensure each location is visited at least once and only on one route


jayroos:
I am trying to build routes that make sure each location is visited at least once and generally only on one route.

The background is we have 37 locations with 3 routes and a central depot. Every location receives a delivery (job) from the depot. We also have shipments taking place between locations where items are picked up at one location and delivered to another.

Ideally, once all locations are visited the vans will return to the depot where undelivered items are unloaded, sorted and loaded into vans for the next day.

If I only have deliveries (jobs) in the input file, every location gets just one visit. However, to ensure that we don't exceed vehicle capacity, I need to also consider the shipments between locations. When I add shipments between the non-depot locations, then stops get dropped.

I have attempted to fix this by giving deliveries from the depot priority 100 and shipments between stops priority 1, but it did not resolve the issue.

I have perused past issues and the documentation, but have not figured out how to resolve this. Any help is appreciated.

jcoupey:
> When I add shipments between the non-depot locations, then stops get dropped.

You mean you get unassigned tasks? That's a normal consequence when you have too many tasks along with constraints: capacity, time windows etc.

I'm not sure I totally get what exactly you think is wrong. Could you share a [standalone problem instance](https://github.com/VROOM-Project/vroom/wiki/Standalone-problem-instance) along with the solution we provide and point out what you'd expect instead (or what is the problem in the existing solution)?

jayroos:
Yes, unassigned tasks. I need the delivery jobs to all get assigned. I'm fine with having pickup and delivery shipments get dropped. I tried to use priorities to ensure the delivery visits would still all be performed, but it didn't seem to have much effect. All locations are visited when it's a delivery from the depot only. When I add in the pickup and drop off shipments then they start dropping off. 

I am attaching two standalone instances (delivery only and delivery and pickup). Both of them have our required time constraint which is more than sufficient for the deliveries alone. I also have a capacity restraint, but I've raised it high enough that it wouldn't ever be reached. I tried removing it, but it seems I have to remove it from all of the jobs/shipments or it won't parse.

[route-matrix.zip](https://github.com/VROOM-Project/vroom/files/7686357/route-matrix.zip)

jcoupey:
[...]

PR #1021, opened 2023-10-24
Priority-oriented local search operator

jcoupey:
## Issue

Attempt at fixing #988 

## Tasks

 - [x] Store forward/backward accumulated priority score at route level
 - [x] Design a move replacing the biggest possible start portion of a route by an unassigned job with higher overall priority
 - [x] Extend that to the end portion of a route
 - [x] Benchmark
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Default radius of 35km for OSRM snapping (#922)
 - Support for URL path in host (#966)
 - Experimental `TSPFix` local search operator (#737)
+- `PriorityReplace` local search operator (#988)
 
 ### Changed
 
--- a/src/algorithms/local_search/local_search.cpp
+++ b/src/algorithms/local_search/local_search.cpp
@@ -452,8 +452,11 @@ void LocalSearch<Route,
 
           assert(fwd_over_rank > 0 || bwd_over_rank > 0);
 
-          if (best_priorities[source] <=
-              std::max(begin_priority_gain, end_priority_gain)) {
+          const auto best_current_priority =
+            std::max(begin_priority_gain, end_priority_gain);
+
+          if (best_current_priority > 0 &&
+              best_priorities[source] <= best_current_priority) {
 #ifdef LOG_LS_OPERATORS
             ++tried_moves[OperatorName::PriorityReplace];
 #endif
@@ -470,9 +473,6 @@ void LocalSearch<Route,
             if (r.is_valid() &&
                 (best_priorities[source] < r.priority_gain() ||
                  (best_priorities[source] == r.priority_gain() &&
-                  // Avoid cycling between replacements with gain
-                  // improvement but zero net priority gain.
-                  r.priority_gain() > 0 &&
                   best_gains[source][source] < r.gain()))) {
               best_priorities[source] = r.priority_gain();
               // This may potentially define a negative value as best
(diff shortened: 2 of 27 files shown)


jcoupey:
## Benchmarking context

Benchmarking this change is quite hard. I do have a few real-life examples that are challenging wrt priority scores and where we produce possibly sub-optimal solutions. On those, this PR tends to generally improve the priority score, the difference being more noticeable for small exploration factors. The instance from #623 is also drastically improved.

Other than that I've been trying to systematically produce custom benchmark instances to confirm the improvement and I found that it's actually quite hard to generate instances where v1.13.0 produces obvious sub-optimal priority scores (that's good news in a way!). What I ended up doing is to generate instances where the priority score is totally at odd with the usual cost criterion, because that's where our approach is more likely to fail reaching good priority scores. 

## Custom instances

The process is as follow:

1. generate random instances with vehicle TW and no job priorities, making sure we don't have enough vehicles to handle everything;
2. solve each instance to spot unassigned jobs (those are the "more expensive ones");
3. go back to the initial instance and set a "priority": 100 value to a random fraction (1%, 2%, 4%, 10%, 20%, 40%, 60%) of those unassigned jobs to generate new instances with priorities.

This way we make sure the "though jobs" get a high priority which makes it especially challenging to reach high priority scores on those instances. In other word, a solving approach only aiming at cost could produce arbitrarily bad solutions wrt priority scores (that is actually what happens when only looking at our heuristic solutions).

## Results

### Overview

With this PR, the priority scores are significantly improved over v1.13.0 on a number of such instances when running with `-x 0`. It also came as a surprise that the number of unassigned jobs is usually vastly reduced, even in some situations where the priority score is even.

More precisely, on those 154 instances the rate of solutions reaching the upper bound for priority score goes from 52.0% for v1.13.0 to 79.2% for 33486e1c.

Rate of solutions with strict improvement with PR by criteria:

Criteria | strictly improved solutions
-- | --
Priority | 39.6%
Assigned jobs | 62.3%
Cost | 55.2%

It's notable that whenever assigned jobs and cost are not improved, it is in most situations because the priority score has increased, resulting in a "more constrained" solution. This makes sense with our criterion ordering.

### Example

Here is an example with only 1 job out of 200 with priority 100 (coordinates [2.21221, 48.30892]).

Solution with v1.13.0 (priority 100, 107 unassigned jobs and cost of 149235)

j_200_v_10_seed_18_prioritized_01_sol


Solution with PR (priority 100, 87 unassigned jobs and cost of 141601)

j_200_v_10_seed_18_prioritized_01_sol

In the PR solution, one route goes further enough to reach that job but the other routes make much more sense, so we're able to assign 20 jobs more within the vehicles TW with an even lower cost.

My interpretation is that the new operator from this PR made it possible to change a good heuristic solution to include the priority job. So we reach a very good solution both cost-wise and in term of assigned jobs, with the optimal priority score. On the other hand with v1.13.0, only one of the heuristic solutions favouring reaching out to far-away jobs was (accidentally) better on priority and thus ended up outputted after the local search. But in that case, the local search did not manage to improve the cost that much, and make room to assign more jobs.

### Impact of exploration level


The impact of the operator is somewhat lower when using `-x 5`, because having a high exploration level is another complementary mean to favor including jobs with priority.

With `-x 5`, the rate of solutions reaching the upper bound for priority is already 76.6% for v1.13.0 and goes up to 81.8% for 33486e1c.

Rate of solutions with strict improvement with PR by criteria:

Criteria | strictly improved solutions
-- | --
Priority | 17.5%
Assigned jobs | 75.3
Cost | 42.2%

So even if the net difference on priority is less significant, we still see an interesting improvement on other criteria. Also notably solutions that are worse in cost are usually better on assigned jobs, and solutions that are worse on assigned jobs (and/or cost) are usually better on priority, which again makes sense with our optimization criterion ordering.

PR #1040, opened 2023-11-17
Fix default time window scaling overflow

kkarbowiak:
## Tasks

 - [ ] review


jcoupey (approved):
Thanks for the PR! I just added a similar change for `scale_*_user_cost`.

Issue #1043, opened 2023-11-21
latest code binary doesn't work with custom matrices input file that worked fine with v1.13.0


answerquest:
For the same input json file with custom matrices included, vroom v1.13.0 is solving properly, but the vroom binary compiled from latest master branch code keeps erroring out

```
# with binary compiled from latest master branch
vroom -i input.json
[Error] Invalid matrix entry.
{"code":2,"error":"Invalid matrix entry."}
```

```
# with binary compiled from v1.13.0
vroom -i input.json
{"code":0,"summary":{"cost":6202,"routes":1,"unassigned":0,"delivery":[4],"amount":[4],"pickup":[0],"setup":0,"service":1200,"duration":6202,"waiting_time":0,"priority":0,"violations":[],"computing_times":{"loading":0,"solving":0}},"unassigned":[],"routes":[{"vehicle":1,"cost":6202,"delivery":[4],"amount":[4],"pickup":[0],"setup":0,"service":1200,"duration":6202,"waiting_time":0,"priority":0,"steps":[{"type":"start","location_index":0,"setup":0,"service":0,"waiting_time":0,"load":[4],"arrival":0,"duration":0,"violations":[]},{"type":"job","location_index":4,"id":4,"setup":0,"service":300,"waiting_time":0,"job":4,"load":[3],"arrival":2074,"duration":2074,"violations":[]},{"type":"job","location_index":3,"id":3,"setup":0,"service":300,"waiting_time":0,"job":3,"load":[2],"arrival":3222,"duration":2922,"violations":[]},{"type":"job","location_index":2,"id":2,"setup":0,"service":300,"waiting_time":0,"job":2,"load":[1],"arrival":3816,"duration":3216,"violations":[]},{"type":"job","location_index":1,"id":1,"setup":0,"service":300,"waiting_time":0,"job":1,"load":[0],"arrival":4802,"duration":3902,"violations":[]},{"type":"end","location_index":0,"setup":0,"service":0,"waiting_time":0,"load":[0],"arrival":7402,"duration":6202,"violations":[]}],"violations":[]}]}
```

Sample input file for testing:
```
[...]

PR #1044, opened 2023-11-22
API.md mention units in matrices

answerquest:
## Issue
https://github.com/VROOM-Project/vroom/issues/1043

Update the documentation to mention that the numbers in the custom matrices need to be integers, and their units (seconds for duration, meters for distance)
 
## Tasks
- [ ] review

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -282,8 +282,8 @@ each vehicle profile. Each matrix is an array of arrays of unsigned
 integers filed under the `profile` key, then under:
 
 - `durations` for a custom travel-time matrix that will be used for
-  all checks against timing constraints;
-- `distances` for a custom distance matrix;
+  all checks against timing constraints (integers, in seconds);
+- `distances` for a custom distance matrix (integers, in meters);
 - `costs` for a custom cost matrix that will be used within all route
   cost evaluations.
 


jcoupey:
I'm not so sure about specifying this here. We have a generic disclaimer at the forefront of the docs:

https://github.com/VROOM-Project/vroom/blob/b9928306a4d44886fb610600ea52b18fb7bb7bf6/docs/API.md?plain=1#L14-L17

If we specify for the matrices, then there is no reason not to specify the unit anytime a duration or a distance shows up in the docs (pretty much everywhere), which somehow defeats the purpose of having a generic description.

Something more efficient in term of feedback for users would be to provide a more explicit error message. If you had read "Matrix value is not an integer." instead of "Invalid matrix entry.", you'd probably have figured out the problem on your own. What do you think?

answerquest:
@jcoupey whoops, thanks for pointing that out, didn't see that earlier. I agree.

And YES, an error message like "Matrix value is not an integer." would help figure out the problem much earlier.

We can close this PR.

jcoupey:
> And YES, an error message like "Matrix value is not an integer." would help figure out the problem much earlier.

Feel free to submit another PR for this!

Issue #1080, opened 2024-04-08
Update OS and compilers in CI


jcoupey:
We use Ubuntu for our CI builds and the next LTS is due by the end of the month so it's a good opportunity to upgrade:

- base image from `ubuntu-22.04` to `ubuntu-24.04`;
- `clang` to version 18;
- `gcc` to version 13.

Potential blockers: 1. I have no idea when the new base will be available in GitHub actions and 2. it's fairly possible that building the upstream OSRM code will raise a handful of new problems with more recent compilers.

jcoupey:
In order to unlock some pending PR/tickets, we need to move on with this.

First clear thing is that we'll never get to cleanly build OSRM v5.27.1 (current latest version) under Ubuntu 24.04 with more recent compilers. On the other hand, building OSRM with gcc 13 and clang 15 is now OK with `master` as of recent fixes.

Now the purpose of our `vroom + libosrm` workflow is to check whether we do not break anything related to using libosrm, or conversely checking that there is no upstream break with our current workflow. Considering that this has been useful in the past, and that our own libosrm integration code does not change much anyway, I think it makes sense that we upgrade OSRM to a latest version, even if it means pinning to a specific `master` commit until there is a new upstream release. I've started toying with this in #1113.

PR #1084, opened 2024-04-11
Refactor/use format

### LINKED ISSUE #1081: Use `std::format`
We have a bunch of warnings in [the sonarcloud report](https://sonarcloud.io/project/issues?resolved=false&rules=cpp%3AS6185&id=VROOM-Project_vroom) about using `std::format`. I did not address this earlier in #984 or #986 simply because the support was missing in compilers at that time.

I suspect newer versions of both gcc and clang would now be OK with the changes so we should dig into this once the CI compiler upgrade in #1080 is completed.

SebMilardo:
## Issue

Fixes #1081 

## Tasks

 - [x] Use std::format instead of concatenating pieces manually
 - [x] review


src/structures/vroom/input/input.cpp
```diff
@@ -186,47 +188,51 @@ void Input::add_job(const Job& job) {
...
+      std::format("Inconsistent shipment amount for pickup {} and delivery {}.",
+                  pickup.id,
+                  delivery.id));
   }
   if (pickup.skills.size() != delivery.skills.size()) {
-    throw InputException("Inconsistent shipment skills for pickup " +
-                         std::to_string(pickup.id) + " and delivery " +
-                         std::to_string(delivery.id) + ".");
+    throw InputException(
+      std::format("Inconsistent shipment skills for for pickup {} and delivery "
```
jcoupey, line 205: Typo here with extra "for".

jcoupey (commented):
Thanks for the PR! Looks great overall, the error messages are much more readable now.

Sonarcloud [is complaining](https://sonarcloud.io/project/issues?id=VROOM-Project_vroom&pullRequest=1084&resolved=false&inNewCodePeriod=true) about not using raw string literals to avoid escaping, but I don't know how this plays along with `std::format`?

SebMilardo:
> Thanks for the PR! Looks great overall, the error messages are much more readable now.
> 
> Sonarcloud [is complaining](https://sonarcloud.io/project/issues?id=VROOM-Project_vroom&pullRequest=1084&resolved=false&inNewCodePeriod=true) about not using raw string literals to avoid escaping, but I don't know how this plays along with `std::format`?

It looks like I have to rewrite them as:

`R"({{"lon":{},"lat":{},"type":"break"}},)"`

and

`R"({{"lon":{},"lat":{}}},)"`

Let me push a commit...

jcoupey:
Thanks for the changes, looks all good to me.

The only weird thing is we'll have to wait to merge since our compiler versions in CI will error on `std::format` currently. Thus there is no point in enabling the CI checks right now either since they're bound to fail. Let's leave this open for now and revisit once #1080 lands.

jcoupey:
We should now be able to let this run through the CI checks after upgrading compiler versions in #1113.

@SebMilardo could you pull the latest master on your fork and `git merge --no-ff master` into your PR branch? I did just that but it looks like I'm not allowed to push to your fork.

SebMilardo:
It should be ok now

src/routing/valhalla_wrapper.cpp
```diff
@@ -37,8 +37,8 @@ std::string ValhallaWrapper::get_matrix_query(
   // List locations.
   std::string all_locations;
   for (auto const& location : locations) {
-    all_locations += "{\"lon\":" + std::to_string(location.lon()) + "," +
-                     "\"lat\":" + std::to_string(location.lat()) + "},";
+    all_locations +=
+      std::format(R"({{"lon":{},"lat":{}}},)", location.lon(), location.lat());
```
jcoupey, line 41: Did you check this produces the same behavior as before? Asking before I don't really get the reason we have to use `{{` and `}}`.

jcoupey (commented):
Thanks for the update, CI builds are OK now! \o/

jcoupey (approved):

PR #1101, opened 2024-04-17
Set compiler for OSRM build

kkarbowiak:
## Tasks

 - [ ] review

### DIFF
--- a/.github/workflows/vroom_libosrm.yml
+++ b/.github/workflows/vroom_libosrm.yml
@@ -17,13 +17,12 @@ env:
   osrm-tag: v5.27.1
 jobs:
   libosrm:
-    strategy:
-      matrix:
-        cxx:
-          - g++-12
-            cc: gcc-12
-          - clang++-15
-            cc: clang-15
+    matrix:
+      cxx:
+        - g++-12
+          cc: gcc-12
+        - clang++-15
+          cc: clang-15
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout repository
--- a/.github/workflows/vroom_libosrm.yml
+++ b/.github/workflows/vroom_libosrm.yml
@@ -19,7 +19,11 @@ jobs:
   libosrm:
     strategy:
       matrix:
-        cxx: ['g++-12', 'clang++-15']
+        cxx:
+          - g++-12
+            cc: gcc-12
+          - clang++-15
+            cc: clang-15
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout repository
@@ -51,7 +55,7 @@ jobs:
         run: |
           mkdir build
           cd build
-          cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=${{ matrix.cxx }}
+          cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} -DCMAKE_C_COMPILER=${{ matrix.cc }}
           cmake --build .
         working-directory: osrm-backend
       - name: Install OSRM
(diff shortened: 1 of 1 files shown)


kkarbowiak:
While current master branch compiles fine using clang-15 on my machine, the v5.27.1 release sadly does not. I guess I forgot about the compilation problems with this project.

kkarbowiak:
Compilation with g++-12 fails for both v5.27.1 and current master.

jcoupey:
Building OSRM with g++-12 works on v5.27.1 when adding `-DCMAKE_CXX_FLAGS="-Wno-array-bounds -Wno-uninitialized"` to the first `cmake ..` command. There are a couple issues related to the error without those flags on the OSRM bugtracker. Long story short this is how it's done on the OSRM CI pipeline [for gcc-12](https://github.com/Project-OSRM/osrm-backend/blob/7ebd21f39e32640e0d40b893d228350f218085ad/.github/workflows/osrm-backend.yml#L224).

I'll try to investigate the clang-15 errors since we'll have to sort them out in some way if we ever want #1080 to happen before another upstream release.

jcoupey:
The clang++-15 errors are fairly simple and have been fixed upstream in `master` in various commits. Porting the changes back to v5.27.1 boils down to applying this patch:

```patch
diff --git a/include/util/range_table.hpp b/include/util/range_table.hpp
index aa1b89f1c..31379ada6 100644
--- a/include/util/range_table.hpp
+++ b/include/util/range_table.hpp
@@ -81,10 +81,10 @@ template <unsigned BLOCK_SIZE, storage::Ownership Ownership> class RangeTable
         unsigned last_length = 0;
         unsigned lengths_prefix_sum = 0;
         unsigned block_idx = 0;
-        unsigned block_counter = 0;
         BlockT block;
 #ifndef BOOST_ASSERT_IS_VOID
         unsigned block_sum = 0;
+        unsigned block_counter = 0;
 #endif
         for (const unsigned l : lengths)
         {
@@ -111,7 +111,9 @@ template <unsigned BLOCK_SIZE, storage::Ownership Ownership> class RangeTable
             if (BLOCK_SIZE == block_idx)
             {
                 diff_blocks.push_back(block);
+#ifndef BOOST_ASSERT_IS_VOID
                 block_counter++;
+#endif
             }
 
             // we can only store strings with length 255
diff --git a/src/contractor/graph_contractor.cpp b/src/contractor/graph_contractor.cpp
index 0b3e87cc2..17a386f11 100644
--- a/src/contractor/graph_contractor.cpp
+++ b/src/contractor/graph_contractor.cpp
@@ -645,7 +645,6 @@ std::vector<bool> contractGraph(ContractorGraph &graph,
 
     const util::XORFastHash<> hash;
 
-    unsigned current_level = 0;
     std::size_t next_renumbering = number_of_nodes * 0.35;
     while (remaining_nodes.size() > number_of_core_nodes)
     {
@@ -761,7 +760,6 @@ std::vector<bool> contractGraph(ContractorGraph &graph,
         remaining_nodes.resize(begin_independent_nodes_idx);
 
         p.PrintStatus(number_of_contracted_nodes);
-        ++current_level;
     }
 
     node_data.Renumber(new_to_old_node_id);
```

Unfortunately the two commits containing those changes also contain other changes in other files (or the same files) so that it is no possible to simply cherry-pick just the above on top of v5.27.1.

What about storing this patch somewhere in our `include` folder and applying in CI? I'm aware this would only be a temporary workaround, but as far as OSRM releases are concerned a temporary workaround may be here to stay...

jcoupey:
Closing here as the scope of this PR has been addressed as part of #1113. Thanks @kkarbowiak for the `include` tip for additional variables in the worfklow.

Issue #657, opened 2022-01-25
Avoid copies of SWAP* insertion ranges


jcoupey:
In the SWAP* operator, we have to check validity for the addition of a range that consist of a portion of a route with the addition of a single job that is either appended at the front or the back of that range.

When sketching the implementation in #580, I couldn't come up with anything more efficient than just copying the initial range to account for both situations with the added job before and after:

https://github.com/VROOM-Project/vroom/blob/ae413af69663c3ce1c67b134f48b647920f97837/src/algorithms/local_search/swap_star_utils.h#L101-L115

This uses `std::copy` for the initial range and a `push_back` either before or after copying and is of course very expensive, especially since we have quite a lot of swap choices options to go through. I suspect this is responsible for most of the computing time devoted to this local search operator.

Now the initial need is "simply" to pass iterators for the candidate range to the various `is_valid_for_*` validity check functions down the line for tests so it feels wrong to have to copy everything for that purpose. On the other hand, I have not found a more efficient way to describe that range and avoid the copies.

Happy to receive any idea or suggestion on that one, maybe the new range library can help here?

jcoupey:
The more I think about this the more I'm convinced this would be a big step forward to easily further reduce computing times. Taking a look at how ranges work, it seems we could achieve the desired effect using `views::concat` from [the range-v3 implementation](https://ericniebler.github.io/range-v3/). This library has been prefiguring the standard for ranges in C++20 but for some reason it looks like `concat` didn't make it to the standard.

On the other hand, we don't really need to concat any arbitrary views, simply being able to iterate over an existing vector + a single value before or after. Probably we can come up with some custom structure owning a reference to the vector and the value and offering the required iteration interface.

kkarbowiak:
@jcoupey do you think this is still relevant and worth pursuing?

Copying some values into a `std::vector` should be very fast, especially for primitive types whare it can boil down to a `memcpy` operation. It might be easier for the compiler to optimise if we used `std::vector::insert` insted of `std::copy`. The only somewhat expensive operation is memory allocation. This could be mitigated in several ways. We could reuse the vector, so once some memory is allocated for one copy, it will be reused for subsequent ones (unless more memory is needed). The vector might need to be a static or thread-local variable. We could also preallocate some memory to minimise the need for subsequent reallocations.

On the other hand I did prepare a small class that behaves like a view allowing iteration over a vector with an additional element in front or at back, but am yet to integrate it with the rest of the code.

[...]

PR #1106, opened 2024-04-26
Copy avoidance in SWAP* insertion ranges

kkarbowiak:
## Issue

This is a draft of proposed change for #657.

The `prepostview` is a template class representing a lightweight view over contents of a vector (or a part of it) with an additional element in front or at back. It has just enough facilities to allow it being used in a range-based for loop.

This is work in progress and needs some polishing before it can be merged, but should work and allow to run tests and benchmarks.

Basically, we are trading a copy + trivial iteration over copied range for no copy but slightly slower iteration (each dereference contains some logic).

## Tasks

 - [ ] ...
 - [ ] review

### DIFF
--- a/src/algorithms/local_search/swap_star_utils.h
+++ b/src/algorithms/local_search/swap_star_utils.h
@@ -14,6 +14,7 @@ All rights reserved (see LICENSE).
 
 #include "algorithms/local_search/top_insertions.h"
 #include "structures/typedefs.h"
+#include "structures/prepostview.h"
 #include "structures/vroom/input/input.h"
 #include "structures/vroom/solution_state.h"
 #include "utils/helpers.h"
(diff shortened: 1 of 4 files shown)


jcoupey:
> I want to look at instance-level profiling/flamegraph to dig a bit deeper before drawing any conclusion.

So I finally got around to doing more thorough profiling. I've compared 176f57a9 (parent on master) and current HEAD of this PR at 463320b3 on several instances.

I've been using perf to monitor the share of `ls::compute_best_swap_star_choice` in the overall computation (this function takes a significant part in `ls::Operator::gain` computations). The change in number of samples and share is not representative of any significant improvement with the PR.

@kkarbowiak I guess my initial intuition from #657 was flawed, much as suggested by [your comment](https://github.com/VROOM-Project/vroom/issues/657#issuecomment-2066515373).

jcoupey:
Closing here as the initial copy-avoidance approach does not seem relevant after all. I'll keep the related ticket open though because it contains other interesting ideas.

jcoupey:
@kkarbowiak thanks for allowing to evaluate the idea with this draft!

Issue #1098, opened 2024-04-17
Mismatch in compilers used for OSRM and VROOM in CI


kkarbowiak:
It looks like the gcc build uses the default g++ to build OSRM (11.4.0) and g++-12 (12.3.0) to build vroom.

For clang build it is g++ 11.4.0 for OSRM and clang-15 for vroom. How this one works I have no idea.

jcoupey:
@kkarbowiak thanks for spotting the problem and reporting. I've ported your suggested changes from #1101 in #1113 as part of the global OS/compiler CI update.

PR #1113, opened 2024-05-21
Move CI to Ubuntu 24.04

### LINKED ISSUE #1080: Update OS and compilers in CI

jcoupey:
## Issue

Fixes #1080 

## Tasks

 - [x] Update Ubuntu base image for CI checks
 - [x] Update compiler versions
 - [x] Debug build errors
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/.github/workflows/vroom.yml
+++ b/.github/workflows/vroom.yml
@@ -16,7 +16,7 @@ jobs:
   vroom:
     strategy:
       matrix:
-        cxx: ['g++-13', 'clang++-18']
+        cxx: [g++-14, clang++-18]
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout repository
--- a/.github/workflows/vroom.yml
+++ b/.github/workflows/vroom.yml
@@ -16,8 +16,8 @@ jobs:
   vroom:
     strategy:
       matrix:
-        cxx: ['g++-12', 'clang++-15']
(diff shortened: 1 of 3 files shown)


jcoupey:
Right, so the full workflow is now OK with gcc 13 after:
- pinning OSRM to current `HEAD` with hard-coded commit SHA in the GA workflow;
- adding ` -DCMAKE_CXX_FLAGS="-Wno-array-bounds` to the OSRM build command.

Our next problem is that we do not really check for clang 18 due to a known issue reported in #1098. I'll try to include the changes suggested in #1101 here in order to try and do everything at once.

My expectation is that we'll have everything running then except the OSRM clang 18 build, which we can then focus on.

jcoupey:
As expected, building OSRM on current `master` fails with clang 18. The good news is that it's OK with clang 16 and 17 which I just tested locally, so the upgrade (and OS update) is still possible even if we have to wait to be using clang 18.

PR #1123, opened 2024-06-10
Solution ruin steps

### LINKED ISSUE #874: Rework solution space jump
Once we've reached a local minima, we try jumping around in the solution space by removing `n` jobs from each route, then re-inserting them heuristically, then rerunning a local search phase. Here `n` depends on the exploration level and current depth stage.

While the overall principle of ruining and re-creating a solution once in a while has proven its worth, the current implementation has major drawbacks.

1. `n` is between 1 and 5. This was deemed enough at the time of implementation while testing on Solomon 100-jobs instances, but does not make sense on much bigger instances. The number of removed tasks should probably be proportional to the instance size.
2. The current process removes exactly the same number of tasks *in all routes*. It would probably be much more efficient to remove "overall best candidates" (to be defined). This way some routes may stay virtually untouched while others could be mostly destroyed.
3. The current way the best candidates per route are found is somehow expensive, especially when applied several times in a row.

Those statements bring more questions than answers, but I really feel we could make a better use of the search depth if we improve the current logic.

### LINKED ISSUE #1124: Log solving profiles
The only debug info we currently have related to the local search process is a count of the number of time each operator is applied. To further analyze the current solving pattern, and the impact of further changes (e.g. #874), it would be useful to log the evolution of each of the parallelized search process:

- current solution indicators;
- time;
- triggering event, such as operator application, job addition, solution R&R etc.

We could then have a few scripts to analyze and plot the data in https://github.com/VROOM-Project/vroom-scripts/.

jcoupey:
## Issue

Fixes #874 

Fixes #1124 

## Tasks

 - [ ] Log LS solving steps
 - [ ] Implement and test various ruining strategies
 - [ ] Benchmark
 - [ ] Update `CHANGELOG.md`
 - [ ] review

### DIFF
--- a/libvroom_examples/libvroom.cpp
+++ b/libvroom_examples/libvroom.cpp
@@ -218,11 +218,11 @@ void run_example_with_osrm() {
   // - jobs 3 and 4 can only be served by vehicle 2
   // - jobs 5 and 6 can be served by either one of the vehicles
 
-  // Solve!
+  // Solve using exploration level as depth and number of searches.
   auto sol =
-    problem_instance.solve(vroom::DEFAULT_EXPLORATION_LEVEL, // Exploration
-                                                             // level.
-                           vroom::DEFAULT_THREADS_NUMBER);   // Use 4 threads.
+    problem_instance.solve(vroom::DEFAULT_EXPLORATION_LEVEL,
+                           vroom::DEFAULT_EXPLORATION_LEVEL,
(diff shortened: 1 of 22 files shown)


jcoupey:
The technical work on #1124 is done but #874 requires more research. As the changes are starting to grow and the debug part could be useful for other stuff, I'm closing this PR and will re-open two independent ones to narrow the scope of each.

Issue #1126, opened 2024-06-13
API documentation on distance information in output


kkarbowiak:
For both total distance in all routes in summary and total route distance for each route, the documentation states that this information is available if `-g` flag is used. However, the information is also provided if custom distance matrix is passed in input.

jcoupey:
Thanks for spotting the miss, the behavior indeed changed when we introduced the ability to provide custom distance matrices in v1.14.

jcoupey:
Done in #1127.

PR #1127, opened 2024-06-13
Add information on distance matrices in input

kkarbowiak:
## Issue #1126 

Proposed wording for information on `distance` entries being available when distance matrices are passed ininput.

I think the change is relevant for both master and the 1.14 release branches.

## Tasks

 - [x] review


jcoupey (approved):
I'll merge to `master`. Although this is indeed relevant for the v1.14 release branch, I don't think it's worth the trouble doing a patch release just for a minor fix in the docs.

Issue #1107, opened 2024-04-29
Evaluating using Boost.JSON


jcoupey:
Boost has its own JSON library, and I've heard that it's easier and more intuitive to use than rapidjson. Switching may not harm performance, provided we trust [their benchmarks](https://www.boost.org/doc/libs/1_84_0/libs/json/doc/html/json/benchmarks.html).

Add to this the fact that rapidjson is painfully ~~unmaintained~~ unversioned, maybe we should consider switching. Of course we'd have to evaluate that in the light of our use-case. The most demanding part for us is probably the parsing of huge matrices spitted out by routing engines.

SebMilardo:
Hi! Could you assign this issue to me?

SebMilardo:
The new library seems to be working. However there are some changes that I'd like to discuss. According to https://stackoverflow.com/questions/68105736/boost-json-serialization-format-boost-1-76-0 the Boost Json library outputs doubles in scientific notation. E.g.: 

```
./vroom -i ../docs/example_1.json -r osrm -a car:routing.openstreetmap.de/routed-car/ -p car:443 
[...]

PR #1129, opened 2024-06-17
Use Boost Json

SebMilardo:
## Issue

#1107 

## Tasks
 - [ ] test performance
 - [ ] review

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "include/rapidjson"]
 	path = include/rapidjson
 	url = https://github.com/Tencent/rapidjson
+[submodule "include/json"]
+	path = include/json
+	url = https://github.com/boostorg/json
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,3 @@
 [submodule "include/rapidjson"]
 	path = include/rapidjson
 	url = https://github.com/Tencent/rapidjson
-[submodule "include/json"]
-	path = include/json
-	url = https://github.com/boostorg/json
--- a/src/makefile
+++ b/src/makefile
@@ -6,7 +6,7 @@
 # Variables.
 CXX ?= g++
 USE_ROUTING ?= true
-CXXFLAGS = -MMD -MP -I. -std=c++20 -Wextra -Wpedantic -Wall -O3 -DASIO_STANDALONE -DUSE_ROUTING=$(USE_ROUTING)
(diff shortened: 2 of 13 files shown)


jcoupey:
I'm really curious to evaluate this change for our use-case. In my experience the main bottleneck for us when it comes to json is parsing huge matrices, either from a file (in case of a custom matrix), or from the network (e.g. `osrm-routed` output). In this situation parsing matrices will dominate the problem loading time reported in `summary.computing_times.loading`.

@SebMilardo did you run any tests so far?

SebMilardo:
Not yet, but I'm planning to run some tests this weekend

jcoupey:
Great! You probably want to test various problem sizes, including instances with several thousands of point to really notice differences. If you're only interested in the loading time and the solving time becomes a pain, you may make things faster by 1. only using TSP instances (the dedicated code scales better) 2. using `-l 0` to stop the search prior to any local search.

SebMilardo:
Bad news, I've started testing the `parse` function in `input_parser` and `rapidjson` is always faster than `boost::json`. I'm using an hand crafted input file with 100000 vehicles and 100000 jobs (a ~20MB .json file) and basically `boost::json` is faster at parsing the string but slower at accessing the parsed objects which resulted in the `parse` function being ~25% slower on average. I'm playing around with options, allocators, error checks, etc to make `boost::json` faster but I also found this FAQ (https://www.boost.org/doc/libs/1_85_0/libs/json/doc/html/json/frequently_asked_questions.html) and this library (https://github.com/simdjson/simdjson). `Simdjson` seems to be way faster than both `rapidjson` and `boost::json` at the cost of creating read only objects. As Vroom uses the Json objects to build its own objects it might be worth a try.

jcoupey:
> Simdjson seems to be way faster than both rapidjson and boost::json at the cost of creating read only objects.

Might be a good option. As you point out: we never modify parsed objects but just read parts of them to populate our own C++ objects. Also the dev around `simdjson` seems to be quite active. My concern here would be the time spent, as you already invested quite some in order to adjust the whole codebase for `boost::json`. Do you think it would be easier/faster/possible to start with a quick benchmark outside VROOM?

SebMilardo:
No problem! I think I can integrate `simdjson` in the `parse` function and compare the results. Now that I have a general understanding of where .json data is used in the codebase it is just a matter of learning how this new library works.

PR #1137, opened 2024-07-12
Only use route requests in plan mode

### LINKED ISSUE #444: Bypass matrix request for plan mode
The `plan` mode as described in #430 recently landed in `master`, replacing the usual solving approach behind `Input::solve` with a call to the new `Input::check` function.

As I focused on the implementation feature-wise, I simply re-used the existing matrix available after a request to the routing engine. As a result there are unnecessary calculations performed on the routing side: as task ordering is provided in input for all routes, we could actually get all the information we need by firing one route request per vehicle instead of a request for the whole matrix.

I'm not exactly sure what would be the most elegant way to do this without adding too much boilerplate. A "lazy" approach would be to not change the client code but add a sparse-filling function that would populate only the required entries in the matrix using route requests. A somewhat cleaner approach could be to change the routing `Wrapper` interface to expose the result of a route request. This response would then be used by the client code in `choose_ETA` to derive all required durations.

Something else to keep in mind: upon generating the solution, we populate the routes with distance values using the exact same route request, so it should actually be run only once and stored somewhere for reuse.

jcoupey:
## Issue

Fixes #444 

## Tasks

 - [x] Expose relevant info from `OsrmRoutedWrapper`
 - [x] Expose relevant info from `ValhallaWrapper`
 - [x] Expose relevant info from `ORSWrapper`
 - [x] Compute sparse matrices from `HttpWrapper`
 - [x] Compute sparse matrices from `LibosrmWrapper`
 - [x] Use sparse matrices from `Input::set_matrices`
 - [x] Make sure to run route requests in parallel
 - [x] Also store route geometries to populate solution
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/src/routing/libosrm_wrapper.cpp
+++ b/src/routing/libosrm_wrapper.cpp
@@ -120,6 +120,62 @@ Matrices LibosrmWrapper::get_matrices(const std::vector<Location>& locs) const {
   return m;
 }
 
+void LibosrmWrapper::update_sparse_matrix(
+  const Id v_id,
+  const std::vector<Location>& route_locs,
+  Matrices& m,
+  std::mutex& matrix_m,
+  std::unordered_map<Id, std::string>& v_id_to_geom,
+  std::mutex& id_to_geom_m) const {
+  // Default options for routing.
+  osrm::RouteParameters params(false, // steps
+                               false, // alternatives
+                               false, // annotations
+                               osrm::RouteParameters::GeometriesType::Polyline,
+                               osrm::RouteParameters::OverviewType::Full,
+                               false // continue_straight
+  );
+  params.coordinates.reserve(route_locs.size());
+
+  for (const auto& loc : route_locs) {
+    assert(loc.has_coordinates());
+    params.coordinates.emplace_back(osrm::util::FloatLongitude({loc.lon()}),
+                                    osrm::util::FloatLatitude({loc.lat()}));
+  }
+
+  osrm::json::Object result;
+  osrm::Status status = _osrm.Route(params, result);
+
+  if (status == osrm::Status::Error) {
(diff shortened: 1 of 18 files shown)


jcoupey:
Storing geometries from first route requests in a `std::unordered_map` based on vehicle id is not a good idea since technically having vehicles with identical ids is allowed.

src/routing/libosrm_wrapper.cpp
```diff
@@ -120,40 +122,85 @@ Matrices LibosrmWrapper::get_matrices(const std::vector<Location>& locs) const {
...
+    m.distances[route_locs[i].index()][route_locs[i + 1].index()] =
+      utils::round<UserDistance>(
+        leg.values["distance"].get<osrm::json::Number>().value);
+  }
+
+  std::scoped_lock<std::mutex> lock(id_to_geom_m);
+  v_id_to_geom.try_emplace(v_id,
+                           std::move(json_route.values["geometry"]
+                                       .get<osrm::json::String>()
+                                       .value));
```
jcoupey, line 187: Storing geometries from first route requests in a `std::unordered_map` based on vehicle id is not a good idea since technically having vehicles with identical ids is allowed.

jcoupey (commented):
Still a UB problem with the edge case of multiple vehicles with same id.

jcoupey:
It's expected that the benefit for this PR grows with the number of involved locations. Benchmarking shows a small benefit for *very small* `plan` requests (think just a few locations). But I observed roughly:

- a 2x speedup on real-life planning examples (routes in the tenths of stops);
- a 10x speedup on custom instances with several hundreds locations (actual gain depends on the number of vehicles).

PR #1177, opened 2024-10-29
Apply heuristics for partial solutions

### LINKED ISSUE #977: Applying heuristics to partial solutions provided in input
Currently when a partial solution is provided in input (using `vehicles.steps`) in solving mode, then we bypass all the heuristic process altogether to initialize the solution with the manual one, if valid. Then as usual it is up to the local search process to try improving that initial solution and put in as much as possible of the jobs unassigned so far.

In particular, this means that instead of doing several parallelized local search, we only apply the local search to that one solution.

This is not ideal in the situation where the provided solution is quite partial, say with only a fraction of initial jobs assigned. The way missing jobs are further assigned by the local search narrows down to a single search path and we miss the diversity of exploration offered by applying multiple heuristic tunings to get multiple initial solutions as we usually do.

Now that we have #837, we could instead:

- set the manual solution as a common starting point;
- apply all heuristics tunings as usual (meaning we fill up the common initial solution in different ways);
- run all local searches on the various resulting initial solutions as usual.

If the partial solution is nearly full, we may end up with almost all initial solutions being identical, but that would not be a problem since we now filter out identical initial solutions before firing the local searches (see #750 and linked stuff).

### LINKED ISSUE #1181: Code deduplication in heuristics
As sonarcloud reports in #1177, there is quite some code duplication across the `basic` and `dynamic` heuristics implementation. We could refactor to simplify the code and reduce duplication with a few helper functions, the main (quite) similar part being the route seeding strategy.

jcoupey:
## Issue

Fixes #977, fixes #1181.

## Tasks

 - [x] Refactor initial solution setup to happen prior to heuristic application
 - [x] Update `basic` heuristic to account for non-empty routes in input
 - [x] Update `dynamic` heuristic to account for non-empty routes in input
 - [x] Move generating unassigned jobs for `basic` and `dynamic` heuristics into a single helper function
 - [x] Move seeding logic for `basic` and `dynamic` heuristics into a single helper function
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/src/algorithms/heuristics/heuristics.cpp
+++ b/src/algorithms/heuristics/heuristics.cpp
@@ -39,7 +39,8 @@ inline void seed_route(const Input& input,
   for (const auto job_rank : unassigned) {
     const auto& current_job = input.jobs[job_rank];
 
-    if (job_not_ok(job_rank) || current_job.type == JOB_TYPE::DELIVERY) {
+    if (!input.vehicle_ok_with_job(v_rank, job_rank) ||
+        current_job.type == JOB_TYPE::DELIVERY || job_not_ok(job_rank)) {
       continue;
     }
 
@@ -412,9 +413,8 @@ Eval basic(const Input& input,
     auto& current_r = routes[v_rank];
 
     if (current_r.empty() && init != INIT::NONE) {
-      auto job_not_ok = [&input, v_rank](const Index job_rank) {
-        return !input.vehicle_ok_with_job(v_rank, job_rank);
-      };
+      // Trivial lambda for no additional job validity constraint.
+      constexpr auto job_not_ok = [](const Index) { return false; };
 
       seed_route(input, current_r, init, evals, unassigned, job_not_ok);
     }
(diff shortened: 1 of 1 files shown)


jcoupey:
I've tested things related to #977 as follows:

- generate optimized routes for a set of random instances;
- remove various fractions of the routes to generate a set of initial solutions that are more or less "complete";
- solve with partial solution in input.

Compared to `master`, this PR yields much better solutions as expected since applying heuristics to partial solutions broadens the search. In particular in many situations, `master` is unable to assign all tasks back when starting with a single local search on the partial solution, while PR does reach to "full" solutions with no unassigned task. When solutions are full on both sides, PR reaches smaller cost solutions.

Also I've generated specific instances to reproduce the embarrassing situations mentioned [here](https://github.com/VROOM-Project/vroom/issues/977#issuecomment-1705154990) in order to confirm that this PR does fix it.

jcoupey:
Now for the refactoring part from #1181, this PR significantly reduces the number of lines and heuristic stuff is now much more readable.

I've confirmed that we still get the exact same executions and searches on the usual benchmarks. The caveat being that I've notice a slight slowdown for the heuristics computing. I've done some profiling but I have no explanation other than the fact that the core heuristic parts are now factored in helper functions. Those seem to be inlined by the compiler but probably the added boilerplate prevent some further optimization that was happening before.

jcoupey:
Turns out the computing times differences I reported above are similar to the differences that can be observed by just switching compiler, so I don't thing there is a specific flaw in the refactoring.

Issue #1162, opened 2024-09-24
Invalid route reached with `PriorityReplace`


jcoupey:
I stumbled upon an instance triggering a segfault with a v1.14 release build. Using a debug build hits a capacity-related assert.

The instance includes a mix of jobs and shipments along with various priority levels and it turns out the problem is caused by an invalid application of the `PriorityReplace` local search operator introduced as part of #1021. The operator splits a valid route at a point where there are still pending deliveries so we end up in an invalid state.

PR #1179, opened 2024-11-04
Add pending delivery checks in PriorityReplace

nir-ben-menachem:
## Issue

Fixes https://github.com/VROOM-Project/vroom/issues/1162

## Tasks

 - [x] Add pending delivery checks to priority replace operator
 - [ ] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Reduce complexity for recreation process (#1155)
 - Refactor `SolutionIndicators` (#1169)
 - Remove amount consistency checks in `parse` in favor of upstream checks in `Input` (#1086)
+- Add pending delivery constraint checks in `cvrp::PriorityReplace::is_valid`
 
 #### CI
 
@@ -33,6 +34,7 @@
 
 - Crash when input is valid JSON but not an object (#1172)
 - Capacity array check consistency (#1086)
+- Invalid route reached with PriorityReplace (#1162)
 
 #### Internals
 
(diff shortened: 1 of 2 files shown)


jcoupey:
Thanks for suggesting a fix! We indeed need to use the existing check for pending deliveries as part of the validity check in `cvrp::PriorityReplace::is_valid`. That will prevent crashes by avoiding picking an invalid move, but we actually need a broader improvement.

The point of the `PriorityReplace` operator is to try to replace as many tasks at the beginning (resp. end) of a route as possible by a single job, as soon as it improves the priority score. So adding the check will only discard invalid moves, but not account for other possible moves that are both valid and still improving in term of priority.

Illustration: say you have a route starting with `P1->D1->P2->D2->P3->...` and the first 5 tasks combined have a priority smaller than the one for a single job we could insert as replacement. Then simply adding a check for pending deliveries after `P3` will rightly tell that cutting the route right after `P3` is invalid. What we need instead is to try replacing a smaller part, e.g. the `P1->D1->P2->D2` part with the single job, which is also a priority score improvement.

nir-ben-menachem:
Thanks for clarifying, I assume that means the same PR should also add pending delivery checks to `fwd_over` and  `bwd_over` evaluation in `run_ls_step` in addition to the current `is_valid` checks ?

jcoupey:
Simply adding another check upstream will again only filter out unwanted moves, not help reaching out to other valid ones.

The current `fwd_last_rank` value as passed to the `PriorityReplace` ctor is the one we'd like to use if valid wrt pending deliveries, but if not we should reach out for the highest possible value below `fwd_last_rank` that is OK for the pending check (if any). So essentially I think once we found the current `fwd_last_rank` candidate we need to add a loop back with new checks. Same symmetric situation with `bwd_first_rank`.

This is starting to be a lot of boilerplate prior to calling the `PriorityReplace` ctor so I wonder if it wouldn't be better to move the responsibility of determining the ranks to the operator itself.

nir-ben-menachem:
After Discussing with @jcoupey, a more comprehensive fix is in the works

PR #1194, opened 2024-11-22
Improve heuristic regrets

### LINKED ISSUE #1193: Improve heuristic choices when using lots of skills
Over time, we've changed things that have subtle impacts on the heuristic choices to build initial solutions. For example #648 indirectly impacts vehicle ordering, #982 impacts regrets computations, etc.

We have examples showing that altogether those changes induced a regression in solution quality (especially number of assigned jobs) for some instances with complex skills setups. See https://github.com/VROOM-Project/vroom/issues/1102#issuecomment-2307312491, I'm opening this ticket to track improving that situation since the initial ticket is somehow mixed with a different problem.

cc @do-me

jcoupey:
## Issue

Fixes #1193 

## Tasks

 - [x] Adjust regrets in `basic` heuristic whenever all values are equal for jobs not doable down the line
 - [x] Determine if `dynamic` heuristic is affected by the same problem
 - [x] Update `dynamic` heuristic if required based on previous point
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 
 #### Core solving
 
+- Solution quality regression when using lots of skills (#1193)
 - Crash due to wrong delivery values in some validity checks (#1164)
 - Crash when input is valid JSON but not an object (#1172)
 - Capacity array check consistency (#1086)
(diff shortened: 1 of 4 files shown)


do-me:
Thanks Julien, I just tested again with the current v1.14-PR vs v1.8. Now, not only is v1.14-PR significantly faster than v1.8 but the results are now consistently better or on par with v1.8!

To be precise: 
- unassigned jobs: no difference, but v1.14-PR is now able to serve all jobs in comparison to v1.14
- km win: v1.14-PR won 3 times against v.1.8; v.1.8 was in no case better than v1.14-PR

Great work!

jcoupey:
@do-me thanks for the feedback!

PR #1196, opened 2024-11-29
Refactor heuristic/LS calling code

### LINKED ISSUE #1188: Remove heuristic synchronisation
Back in #750 I introduced a mechanism to deduplicate solutions after the heuristic phase. The rationale was to avoid performing the exact same local search on identical solutions across different threads. While this is interesting in theory, there are drawbacks:

- the code is much more complex with two multi-threading phases and a deduplication phase;
- it requires a synchronization point across heuristic threads, which means "fast" heuristic searches are idle before all have completed.

This mechanism may reduce overall computations, but will in most situation not improve completion time, and could even harm it. Another consideration is that identical heuristic solutions happen quite a lot in small/simple benchmark instances (e.g. Solomon C class) due to the nature of the instances and the costs (rounded euclidean distances), but are pretty rare in a real-life setup with a significant number of locations and travel times to the second.

All this to say I think this is not worth the trouble and we should refactor that bit.

jcoupey:
## Issue

Fixes #1188 

## Tasks

 - [x] Move LS stuff into same thread context after heuristic
 - [x] Add heuristic solution dedup mechanism
 - [x] Update LS logging
 - [x] Refactor huge solving lambda
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 #### Internals
 
 - Bypass matrix request in `plan` mode (#444)
+- Account for heuristic time in timeout (#1196)
 - Refactor heuristics to reduce code duplication (#1181)
 - Refactor `Matrix` template class (#1089)
 - Refactor to use `std::format` whenever possible (#1081)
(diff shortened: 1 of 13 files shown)


jcoupey:
The good part with the implementation in this PR is that we're still able to do search deduplication. Also now that heuristic + local search are run into the same context, we can account for the heuristic time compared to the global timeout value.

PR #1197, opened 2024-12-02
Provide backward-compatible `Input::solve`

jcoupey:
## Issue

The signature for `Input::solve` changed to introduce debug options, but this unnecessarily exposes internal variable choices. We should restore the old signature for out-of-the-box exploration level for both C++ and downstream pyvroom compatibility, see https://github.com/VROOM-Project/pyvroom/pull/121#discussion_r1865424722.

## Tasks

 - [x] Add `Input::solve` overload taking exploration level as parameter
 - [x] Expose definition of depth and number of searches as helper functions for consistency
 - [x] Refactor command-line parameters handling
 - [x] review


jonathf (approved):
This looks good.

PR #1210, opened 2025-01-16
Fix `PriorityReplace`

### LINKED ISSUE #1162: Invalid route reached with `PriorityReplace`

jcoupey:
## Issue

Fixes #1162 

## Tasks

 - [x] Add checks for pending deliveries for `PriorityReplace`
 - [x] If start route replacement split has a pending delivery problem, try finding a smaller start route without splitting problem
 - [x] If end route replacement split has a pending delivery problem, try finding a smaller end route without splitting problem
 - [x] Account for number of tasks after applying to discriminate between start or end replacement if both are valid
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 #### Core solving
 
 - Solution quality regression when using lots of skills (#1193)
+- Invalid route reached by `PriorityReplace` (#1162)
 - Crash due to wrong delivery values in some validity checks (#1164)
 - Crash when input is valid JSON but not an object (#1172)
 - Capacity array check consistency (#1086)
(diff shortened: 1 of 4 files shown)


jcoupey:
Fixes #1212

Issue #336, opened 2020-05-02
Setup/service time per vehicle type


jensbe74:
Hi!  Thank you for the great project!
We use VRoom to plan fitters.  It would be perfect if we could also specify a level of performance for the vehicle. 
Example: The service job has a duration of 15 minutes, for example.  The driver of Vehicle 1 needs 100% = 15 min because he works normally.  However, the driver of Vehicle 2 takes longer and only has an efficiency level of 70% = 19.5 min.  He is new at his Job and not so well incorporated. 
Of course, someone can also work faster than the average (15min) and would then have, for example, 120%. 
Not everyone works as fast as the other and so it can be planned optimally. 
Can this feature be integrated?

jcoupey:
Thanks for your feedback and input, I see your point here and this would probably be useful in several situations.

Technically speaking, I think this is doable without changing the way the overall solving approach works. We'd have to store a rate multiplier at the `vehicle` level and apply that whenever we need to evaluate time spent at a job (both for validity checks or when actually modifying a route).

The possible complication I foresee is that when evaluating potential gains related to small route changes, we currently only take into account travel time costs, because service times are, well, constants so far. Now it may require to toy with the very notion of "gain" if we need to distinguish a situation where allocating some jobs to a "fast" worker is better because of the reduced service time (e.g. in the long run it allows to fit more jobs in).

jcoupey:
Another practical downside would be that people may then complain that a single factor at vehicle level is not enough, and would want to be able to apply a different rate based on the job type. This would come down to defining a **matrix** of vehicles/jobs service times, which I think would be fairly unpractical/unreasonable in term of API bloating.

jensbe74:
Of course you can expand the topic.  But I think one factor per vehicle would be a good approach.  I think adding the service time to the travel time would be the right way

leobudima:
I'm interested in working on this and submitting a PR. To be honest (as mentioned in 358 which unfortunately got de-prioritized for now, but looking to pick it up when possible), the environment (C++ etc) is pretty unfamiliar to me, as well as most of the inner workings of the codebase, but I was hoping, @jcoupey, that you would be so kind (as usual) to potentially give some initial pointers to get started? 

Thank you again for the amazing project and support you're providing here - really looking forward to contributing.

jcoupey:
>For the first iteration, if your opinion is that the initial solution only affecting travel times would be much simpler, I would go this >route, and add a simple per-vehicle factor,

[...]

PR #1216, opened 2025-02-04
Draft: service_factor (job performance rate)

sjoerddal-pti:
## Issue
Addresses https://github.com/VROOM-Project/vroom/issues/336 (job performance rate). 

This PR allows setting per-vehicle service_factor by which job service times are scaled. The repo in its current state produces valid solutions, however, as is mentioned by @jcoupey on the original issue, it currently doesn't have any smarts to prefer vehicles that have a more favourable service_factor. I'm looking to brainstorm ideas here to flesh this out a little more.

I figure this would require expanding the CostWrapper/Eval constructs with service_factor. I wonder if simply dividing the final cost by service_factor could be a solution as this would nudge all heuristics towards vehicles with more favourable service_factors but I fear this would too coarse of a way to do it, and it doesn't account for the differing service times it'd yield.

Expanding the whole Eval/CostWrapper setup with service_factor and requiring eval'ing those against a job index as well sounds like it'd be the way to go here. I'm looking for ways of getting this done without impacting performance. I figure this can be done similar to this PR's setup with vehicle_service, so some sort of matrix/array we'd be able to index into based on vehicle/job and pre-calculating all that before the solve starts, so there'd be no unneeded math and we can stick to integer arithmetic.

I'd love any feedback anyone's able to give me on this and work this out further, so please let me know!

## Tasks

 - [ ] Update heuristics
 - [ ] review


jcoupey:
@sjoerddal-pti thanks for starting to work on this and for submitting a PR. Sorry about the response delay here, took me some time since I've been refining a different approach to solving this issue feature-wise.

I've included explanations on the proposed API changes [in the original ticket](https://github.com/VROOM-Project/vroom/issues/336#issuecomment-2751489380), and started #1238.

sjoerddal-pti:
@jcoupey I see. The new API proposal makes a lot of sense, and seems a lot more future-proof to me. Let me know if there's anything I can do to help test. We're currently using the version on this PR, which gives us workable, albeit subobtimal solutions for the time being. I think it's fine to close this PR in favour of your version and build off of that.

jcoupey:
Closing in favor of #1238.

PR #1222, opened 2025-02-11
Use cost bound in heuristic

### LINKED ISSUE #1219: Heuristics speed-up
Solving time spent during the heuristic process is usually quite low compared to overall solving time. It can nonetheless become very long when solving huge problem instances, especially with few constraints.

For a given vehicle/route construction in progress, what we do is loop through all unassigned jobs to evaluate insertion cost (actually a mix of actual cost and regret ponderation) *at all possible route ranks*, then pick the lowest cost. Looping through the existing route for all unassigned jobs means the process does not scale very well with very long vehicle routes (e.g. ~500 jobs per route and thousands of jobs).

I think we could come up with ways to maintain some data per route and unassigned job that would yield an easy lower bound of the insertion cost. This way whenever we have existing valid insertion candidates with low enough insertion cost, we could cut out the *at all possible route ranks* part for many unassigned candidates.

jcoupey:
## Issue

Fixes #1219

## Tasks

 - [x] Store and update relevant data to compute the bound
 - [x] Avoid going through whole route for rank-based insertion tests based on bound
 - [x] Same but for pickup and delivery
 - [ ] ~~Try ordering unassigned jobs to increase bound efficiency~~
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 #### Internals
 
 - Bypass matrix request in `plan` mode (#444)
+- Heuristics speed-up (#1219)
 - Account for heuristic time in timeout (#1196)
 - Refactor heuristics to reduce code duplication (#1181)
 - Refactor `Matrix` template class (#1089)
--- a/src/algorithms/heuristics/heuristics.cpp
+++ b/src/algorithms/heuristics/heuristics.cpp
@@ -189,12 +189,31 @@ template <class Route> struct UnassignedCosts {
                                min_unassigned_to_route[j] - max_edge_cost);
   }
 
-  void update_costs(const Input& input,
-                    Route& route,
-                    std::set<Index>& unassigned,
-                    Index inserted_index) {
-    max_edge_cost = utils::max_edge_eval(input, vehicle, route.route).cost;
+  double get_pd_insertion_lower_bound(const Input& input, Index p) {
+    assert(input.jobs[p].type == JOB_TYPE::PICKUP);
 
+    // Situation where pickup and delivery are not inserted in a row.
+    const auto apart_insertion = static_cast<double>(
+      min_route_to_unassigned[p] + min_unassigned_to_route[p] +
+      min_route_to_unassigned[p + 1] + min_unassigned_to_route[p + 1] -
+      2 * max_edge_cost);
+
+    // Situation where delivery is inserted next to the pickup.
+    const auto next_insertion = static_cast<double>(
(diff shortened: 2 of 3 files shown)


jcoupey:
My attempts at ordering unassigned jobs to favor cutting branches were void: we'd need to maintain an extra data structure + pay the re-ordering price from time to time, which in my tests destroys the benefit of using the bound with unsorted data.

jcoupey:
## Benchmarking setup for job-based instances

### Instances

In order to measure the impact of this PR, I've generated random instances with sizes ranging from 200 to 5000 jobs. Since route length is a major factor in complexity, I've generated several instances per number of jobs to match typical sizes: S (10 jobs per route), M (50 jobs per route), L (100 jobs per route), XL (250 jobs per route) and XXL (500 jobs per route).

### Process

For all those instances I'm solving *only with the heuristics* (no local search) with both current `master` at 99058156 and PR at 31476a73. We're interested in the solving time ratio.

### Results

**Solving speedup between `master` and PR depending on instance and route sizes.**

Jobs | S | M | L | XL | XXL
-- | -- | -- | -- | -- | --
200 | 1.50 | 1.71 | 1.58 |   |
500 | 1.43 | 2.09 | 1.87 | 1.64 |
1000 | 1.86 | 1.73 | 1.82 | 1.71 | 1.69
2000 | 1.68 | 2.65 | 2.36 | 2.31 | 1.86
3000 | 1.92 | 2.99 | 2.61 | 2.68 | 1.84
4000 | 1.93 | 3.02 | 2.53 | 2.59 | 1.88
5000 | 2.04 | 3.25 | 3.04 | 2.48 | 2.02

If we look at total computing time, the speedup is 2.17, meaning PR is more than 2x faster than `master`. Again that is only on heuristic time, which may be a small part of overall computing time is many situations.

jcoupey:
## Benchmarking setup for shipment-based instances

Using the usual Li&Lim benchmark (all classes with instance sizes from 100 to 1000), again solving *only with the heuristics* (no local search) with both current `master` at 99058156 and PR at 31476a73.

**Average computing times (ms) and speedup**

  | 99058156 | 31476a73 | Speedup
-- | -- | -- | --
100 | 5 | 4 | 1.25
200 | 25 | 21 | 1.19
400 | 109 | 85 | 1.28
600 | 291 | 201 | 1.45
800 | 599 | 372 | 1.61
1000 | 984 | 581 | 1.69

PR #1262, opened 2025-05-28
Extend addition candidates after `UnassignedExchange` move

### LINKED ISSUE #1258: Unassigned job with empty compatible vehicle
I've stumbled into an obviously suboptimal solution on a real-life instance where:

- 1 job ends up unassigned;
- there is an empty (unused) vehicle that could serve it.

After investigating, it turns out that the unassigned job has been dropped at some point by an `UnassignedExchange` move during one of the local search phases. This happens in the middle of a rearranging sequence allowing another route to re-order and insert other unassigned jobs. The core problem is that the job discarded in this `UnassignedExchange` operation is never tested against other vehicle candidates than the one involved into the move. As a result it is never inserted in the available empty vehicle.

That is kind of an edge case in the sense that on top of the above chain of events, we need that exact solution to be the best available in term of number of assigned task for the problem to show up, i.e. finding a "full" solution is hard enough to not happen through other parallel searches.

jcoupey:
## Issue

Fixes #1258 

## Tasks

 - [x] Refactor how `SolutionState` update calls are performed in various places during local search
 - [x] Add all compatible vehicles as candidates for job addition after `UnassignedExchange` move
 - [x] Update `CHANGELOG.md`
 - [x] review

### DIFF
--- a/src/problems/cvrp/operators/unassigned_exchange.cpp
+++ b/src/problems/cvrp/operators/unassigned_exchange.cpp
@@ -7,9 +7,6 @@ All rights reserved (see LICENSE).
 
 */
 
-#include <algorithm>
-#include <ranges>
-
 #include "problems/cvrp/operators/unassigned_exchange.h"
 #include "utils/helpers.h"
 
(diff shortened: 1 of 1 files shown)


saurabh119:
Generating the union of compatible vehicles would be redundant since there is already a compatibility check present in Local Search's `try_job_additions`: https://github.com/VROOM-Project/vroom/blob/8b85c74fe0c115c6c065ccc94fc1565de68eb1c7/src/algorithms/local_search/insertion_search.h#L33-L45

So, passing all vehicles would be more efficient than calculating the compatible ones.

PR #1267, opened 2025-06-10
Simplify gain evaluation when moving parts of routes in operators

### LINKED ISSUE #1266: Refactor operator gain evaluation
I started to work on some of the changes required for operator gain evaluation in  #1252, whose purpose is to add a service time component to the costs. Doing so with the current state of the codebase would require to duplicate the "new" service time logic in many places, typically in all the `*::compute_gain` functions across local search operators. This would be:

- error prone;
- breaking the DRY principle;
- a maintenance nightmare.

This situation is a hint that we could do better. Actually the cost evaluation is *already currently* repeating a lot of similar logic across operators. If this were refactored so that the various `compute_gain` functions would call a unified interface, then:

- maintenance of the current cost evaluation code would be made much easier;
- future changes like the one required for #1252 would be much more straightforward.

This would be in a way similar to what we do with `is_valid_addition_for_tw` that operates in a generic manner on a route with a replaced range and an inserted range. The difference here is that we should keep things happening in constant time, so re-use various precomputed things from `SolutionState`, and not go through the inserted ranges.

jcoupey:
## Issue

Fixes #1266 

## Tasks

 - [x] Implement relevant high-level cost computation functions
 - [x] Assert that custom evaluation costs can be replaced by function calls
 - [x] Replace existing evaluation cost with high-level function calls
 - [x] review

### DIFF
--- a/src/problems/cvrp/operators/or_opt.cpp
+++ b/src/problems/cvrp/operators/or_opt.cpp
@@ -136,7 +136,15 @@ Eval OrOpt::gain_upper_bound() {
                                                       s_rank,
                                                       s_rank + 2));
 
-  // TODO compute _reversed_t_gain
+  assert(_reversed_t_gain == utils::addition_cost_delta(_input,
+                                                        _sol_state,
+                                                        target,
+                                                        t_rank,
+                                                        t_rank,
+                                                        source,
+                                                        s_rank,
+                                                        s_rank + 2,
+                                                        REVERSED_INSERTION));
 
   _gain_upper_bound_computed = true;
 
--- a/src/problems/cvrp/operators/or_opt.cpp
+++ b/src/problems/cvrp/operators/or_opt.cpp
@@ -41,110 +41,26 @@ OrOpt::OrOpt(const Input& input,
 }
 
 Eval OrOpt::gain_upper_bound() {
-  const auto& s_v = _input.vehicles[s_vehicle];
-  const auto& t_v = _input.vehicles[t_vehicle];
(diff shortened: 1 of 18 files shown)


jcoupey:
After some testing, it turns out that moving to the most simple interface has some drawbacks in term of efficiency. For example the previous `in_place_delta_cost` evaluation is very straightforward and efficient because it relies on the fact that we're exactly swapping one job with another. Switching that to a generic "replace that range by that other range, but actually the ranges are just one job" adds quite some boilerplate to the code that is run in hot loops.

We'll probably have to figure out some compromise between code simplicity and efficiency.

src/problems/cvrp/operators/or_opt.cpp
```diff
@@ -40,88 +41,26 @@ OrOpt::OrOpt(const Input& input,
...
+
+  _reversed_t_gain = utils::addition_cost_delta(_input,
+                                                _sol_state,
+                                                target,
+                                                t_rank,
+                                                t_rank,
+                                                source,
+                                                s_rank,
+                                                s_rank + 2,
+                                                REVERSED_INSERTION);
```
jcoupey, line 63: This is typical of all operators potentially reversing some edges (`OrOpt`, `{Mixed,Cross}Exchange` and the `Intra*` versions): we now have two calls to `addition_cost_delta` but basically those two call we mostly perform the same calculation except for the small portion evaluating new inserted cost, either reversed or not.

This is less efficient than the previous implementation where everything was evaluated "manually" here, but both cost versions where computed in one go.

Issue #703, opened 2022-04-26
Relative time window for shipments


jwoyame:
This is a feature suggestion.

First: Vroom is excellent, very well written. Thank you especially to @jcoupey for your efforts.

## Use case

Currently, you can set `time_windows` for both pickup and drop-off jobs in a shipment. But suppose you wanted to deliver a pizza within 30 minutes.

Example: A customer wants her pizza around 5pm. You set a pick-up window of [4:00, 4:45], but you don't want the driver to have it in the car for more than 30 minutes. For that reason you have to set a drop-off window too, perhaps [4:30, 5:15]. But that leaves the possibility for a solution where the pickup happens at 4:00 and the drop-off occurs at 5:15, leaving the pizza in the car for over an hour.

That might seem like a rare need, but there are practical cases where a pickup or a drop-off could happen any time of day, but the _difference_ between the two jobs must be constrained. Another case would be an Uber Pool ride, where someone wants to be picked up between 7am and 8am but doesn't want to ride around for the entire day before reaching their drop-off 😀 

## Solution

The problem exists because there is no way to lock one job to another one's time margin throughout the optimization process. I propose a new `time_gap` property for pickup and drop-off jobs.

For the pizza example, the pickup's `time_windows` would have [4:00, 4:45], and the drop-off would have a `time_gap` of 30. As the pickup's `earliest` and `latest` is dynamically updated, the drop-off job's `latest` is constrained to the pickup's `earliest` + `time_gap`. The existing logic will be able to protect the drop-off from being pushed out later.

This could work in the other direction, where the `time_gap` can be put on the pick-up, and the drop-off's `earliest` and `latest` will serve as the anchor. It should also work even if `time_windows` are not assigned.

Looking at `tw_route.cpp`, this seems like it would not be too big of a change. Since it is still window-based, it would not require the entire route to be compressed to exact ETAs first.

Thank you for any comments or further refinements.

ThomasRambaud:
Hi @jwoyame,
To bypass your problem, couldn't you just specifiy multiple time windows with 5 min steps?
[4:00, 4:30]
[4:05, 4:35]
[4:10, 4:40]
[4:15, 4:45]

jcoupey:
I definitely get your point and how this could be useful in some situations so I'm happy to flag this as a feature request. But honestly I don't see an easy way to make that happen any time soon.

[...]

PR #1275, opened 2025-06-27
Feature/cargo lifetime

igor-semenov:
# Cargo Lifetime Constraints for Perishable Goods

  This feature adds support for time-sensitive cargo with expiration constraints to VROOM, enabling optimization of routes for perishable goods logistics.

  🎯 Feature Overview

  Adds max_lifetime constraints to jobs and shipments, allowing VROOM to:
  - Track cargo age from pickup to delivery
  - Detect and report lifetime violations
  - Prioritize time-sensitive deliveries in route optimization
  - Support perishable goods logistics use cases

  📋 Implementation Tasks

  - Add max_lifetime field to Job structure with constraint methods
  - Add cargo tracking fields to Step structure (pickup_time, cargo_age, etc.)
  - Add lifetime statistics to Route structure (max_cargo_age, violations)
  - Extend violation system with LIFETIME violation type
  - Update JSON I/O to parse and output lifetime information
  - Integrate lifetime constraints into heuristics and validation algorithms
  - Add lifetime checking to local search operators
  - Add lifetime-aware cost penalties for time-sensitive jobs
  - Update API documentation with lifetime constraint fields
  - Add test cases for cargo lifetime constraints
  - Verify compatibility with both routing and non-routing builds

  🔧 Usage

```
{
  "jobs": [
    {
      "id": 1,
      "description": "Perishable delivery - 2 hour lifetime",
      "location_index": 1,
      "delivery": [10],
      "max_lifetime": 7200
    }
  ],
  "shipments": [
    {
      "pickup": { "location_index": 2 },
      "delivery": { "location_index": 3 },
      "amount": [5],
      "max_lifetime": 3600
    }
  ]
}
```

  ✅ Testing

  - Successfully builds with and without routing support
  - Test cases validate lifetime constraint handling
  - Route optimization respects cargo expiration times
  - Proper violation detection and reporting

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -70,6 +70,7 @@ A `job` object has the following properties:
 | [`skills`] | an array of integers defining mandatory skills |
 | [`priority`] | an integer in the `[0, 100]` range describing priority level (defaults to 0) |
 | [`time_windows`] | an array of `time_window` objects describing valid slots for job service start |
+| [`max_lifetime`] | maximum duration (in seconds) that cargo can remain in the system before expiring |
 
 An error is reported if two `job` objects have the same `id`.
 
@@ -84,6 +85,7 @@ A `shipment` object has the following properties:
 | [`amount`] | an array of integers describing multidimensional quantities |
 | [`skills`] | an array of integers defining mandatory skills |
 | [`priority`] | an integer in the `[0, 100]` range describing priority level (defaults to 0) |
+| [`max_lifetime`] | maximum duration (in seconds) between pickup and delivery for perishable goods |
 
 A `shipment_step` is similar to a `job` object (expect for shared keys already present in `shipment`):
 
@@ -403,6 +405,8 @@ A `route` object has the following properties:
 | [`description`] | vehicle description, if provided in input |
 | [`geometry`]* | polyline encoded route geometry |
 | [`distance`]** | total route distance |
+| [`max_cargo_age`] | maximum cargo age encountered in this route (in seconds) |
+| [`lifetime_violations`] | number of steps with expired cargo in this route |
 
 *: provided when using the `-g` flag.
 **: provided when using the `-g` flag or passing distance matrices in input.
@@ -427,6 +431,10 @@ A `step` object has the following properties:
 | ~~[`job`]~~ | ~~id of the job performed at this step, only provided if `type` value is `job`~~ |
 | [`load`] | vehicle load after step completion (with capacity constraints) |
 | [`distance`]* | traveled distance upon arrival at this step |
+| [`cargo_pickup_time`] | time when cargo was picked up (for delivery steps with lifetime constraints) |
+| [`cargo_age`] | age of cargo at this step (for delivery steps with lifetime constraints) |
+| [`max_cargo_lifetime`] | maximum allowed lifetime for cargo at this step |
+| [`has_expired_cargo`] | true if cargo has exceeded its lifetime at this step |
 
 *: provided when using the `-g` flag.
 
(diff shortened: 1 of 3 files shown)


igor-semenov:
> Thanks for sharing you code. In order to first get a high-level picture of how you approach this, my broad understanding is:
> 
> 1. you're adding a cost penalty to some cost evaluation functions (e.g. `addition_cost`) based on a job `max_lifetime` value, hoping this will hint the search toward solutions with less "lifetime violations";
> 2. when presenting the solution in output, you check for actual lifetime violations based on arrival times, then add matching violations in output.
> 
> Is that a good summary?

Your Understanding is Spot-On

Part 1: Search-Time Cost Penalties
- Addition cost functions (like addition_cost) add penalties for jobs with max_lifetime values
- Higher penalties for shorter remaining lifetimes (e.g., 100x penalty for <1hr, 50x for <2hr)
- This guides the search algorithm toward solutions that prioritize time-sensitive cargo
- Keeps the search computationally efficient by avoiding hard constraint enforcement

Part 2: Output-Time Violation Detection
- After route construction, choose_ETA function validates actual arrival times
- Calculates precise cargo age: cargo_age = arrival_time - pickup_time
- If cargo_age > max_lifetime, adds VIOLATION::LIFETIME to the solution
- Reports violations in JSON output with detailed timing information

jcoupey:
Thanks for the follow-up. While this may work on some situations from your own use-case, I can see several limitations for this approach to be used globally.

1.  You're introducing a soft constraint. Supplying a `max_lifetime` in input does not give any guarantee at all of the lifetime being respected. Yet there is no real control on the "additional soft cost" introduced (see more below).
2.  Another way of seing the above is that users could now have non-empty violations in `solve` mode, which is a breaking change.
3.  Penalty choices would not work out-of-the-box in all situations:
- values are very add-hoc, probably tuned on your situation, but would not necessarilly work for different use-cases;
- the cost unit itself may be inconsistent with units users have in place via use of `vehicle.costs` ponderation values.
4. When applying penalties *during the search*, you're using the job `max_lifetime` value, not the actual remaining time based on real timing for route steps. That somehow makes sense since I don't think you can easily compute the actual remaining lifetime during the search without drastically increasing the algorithmic complexity. But the consequence is that the ponderation is hinting toward including jobs with certain lifetime values (a bit like the current priority scheme), but it is not hinting toward including those jobs at a position that would minimize lifetime violations.
5. Modifying `addition_cost` is just the tip of the iceberg here: it may be most of what you need to modify for the heuristic solution building part or a simple local search operator such as `Relocate`, but if you look at [some operators examples](https://youtu.be/HHkwYvFMRiI?t=576), you'll see that gain evaluation in local search is more complicated than that (see also #1266 on this).

igor-semenov:
Thank you very much for your feedback and explanations!  
The shortcomings in my solution are due to my limited understanding of the project's architecture.
Could you please recommend a high-level outline for a proper solution to this task, so I can try to approach it in a way that aligns with the project's architectural principles?
And you're absolutely right: the lifetime constraints should be strict — I just couldn't immediately figure out the correct way to enforce them.

jcoupey:
The problem is that I don't think there is a "proper solution to this task" with the current solving approach. That is implementing this would require to change the internals in a way that would significantly impact algorithmic complexity and thus computing times.

Summarizing the problem: during the local search phase, we do countless checks for time windows validity for many potential solution changes. This happens efficiently because we store earliest and latest dates for all tasks in all routes, and update that information upon any route modification.

A consequence of this is that we currently do not track actual arrival times for steps while searching (only time margin). While we could theoretically decide on actual arrival times upon all route modifications (just as we do when presenting the route in output), we could not decide on lifetime validity anyway without adding traversal of additional route portions.

This subject has also been discussed specifically for `shipment` objects in #703.

jcoupey:
Closing here as I don't think we can properly achieve the initial goal of this PR with the current solving approach, see above comments.

Issue #1274, opened 2025-06-26
Avoid allocating moved jobs ranges too early


jcoupey:
Some local search operators require to store a range of "moved jobs" that is basically a `std::vector<Index>`. It is used to conveniently check validity for various constraints such as capacity, TW, etc. This concerns all five `Intra*` moves and `UnassignedExchange`.

Now the usual process for such moves is to first test for potential gain because it's faster, then only once we see a potential improvement that is better than the "current one", then we check for validity. It means that the `_moved_jobs` vectors are only used in a very small number of occurrences, while they are always created as memory allocation and vector populating happens in the operator ctor.

I'd like to evaluate the gain when avoiding unnecessary allocations altogether: move the `_moved_jobs` stuff away from the ctor into a dedicated function that would be called only upon testing validity.

jcoupey:
Turns out this idea does not have any significant impact on computing times, see https://github.com/VROOM-Project/vroom/pull/1278#issuecomment-3409858320.

PR #1278, opened 2025-07-18
Refactor/avoid allocating moved jobs ranges too early

apidemy:
## Issue

https://github.com/VROOM-Project/vroom/issues/1274

## Tasks

 - [ ] Refactoring: Avoid allocating moved jobs ranges too early

### DIFF
--- a/src/problems/cvrp/operators/intra_cross_exchange.cpp
+++ b/src/problems/cvrp/operators/intra_cross_exchange.cpp
@@ -32,7 +32,6 @@ IntraCrossExchange::IntraCrossExchange(const Input& input,
     // check_t_reverse are false.
     check_s_reverse(check_s_reverse),
     check_t_reverse(check_t_reverse),
-    _moved_jobs(t_rank - s_rank + 2),
     _first_rank(s_rank),
     _last_rank(t_rank + 2),
     _delivery(source.delivery_in_range(_first_rank, _last_rank)) {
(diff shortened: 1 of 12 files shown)


src/problems/cvrp/operators/intra_cross_exchange.cpp
```diff
@@ -57,6 +56,11 @@ IntraCrossExchange::IntraCrossExchange(const Input& input,
           _input.jobs[this->t_route[t_rank + 1]].type == JOB_TYPE::DELIVERY &&
           !check_t_reverse &&
           _sol_state.matching_delivery_rank[t_vehicle][t_rank] == t_rank + 1));
+}
+
+void IntraCrossExchange::prepare_moved_jobs() {
+  _moved_jobs.clear();
```
jcoupey, line 62: I'm not sure we really need to `clear` here as the vector has been initialized empty by default in the constructor?

jcoupey (commented):
Thanks for giving this a go!

Did you run some benchmarking to see whether/how those changes impact computing times?

src/problems/cvrp/operators/intra_cross_exchange.cpp
shahab-tplus: For running benchmarks, I am facing memory issues but trying to run some of them.

And for the vectors yes, they are default-initialized(empty vector) so I removed it.

jcoupey:
Thanks for reporting, I've also been running some benchmarking on my side on various CVRP and VRPTW instances. Turns out the difference is always within the measuring error margin, i.e. there is not a more significant difference between `master` and PR than between different runs from the same build.

I've also tried to move the `_moved_jobs` allocation back into the constructor: providing the vector size in ctor but keep filling it in `prepare_moved_jobs` without having to call `resize`. I was wondering if having two different allocations phases (in ctor and in `prepare_moved_jobs`) could be detrimental. Same results: nothing significant either.

I'll simply close this PR and the associated ticket as not impactful. @apidemy thanks again for taking the time to give this a go, that was the only way to figure out whether the idea from #1274 would make an impact or not.

Issue #1218, opened 2025-02-06
Parallel route requests


jcoupey:
Currently whenever we need to populate the routes geometry, we simply loop through routes in the solution and call `Wrapper::add_geometry`:

https://github.com/VROOM-Project/vroom/blob/9905815687607d675b49fc7faf09a309f69460f9/src/structures/vroom/input/input.cpp#L1188-L1198

This means route requests to the underlying routing engine are done sequentially, while we could send them concurrently.

Probably the speedup we'd get would be very small, and surely almost unnoticeable for most instances: getting the matrices and solving is usually much much longer than the routing part. Still it would be nice to try it out.

Note: matrix computing is already parallelized across profiles (`Input::set_matrices` calls `get_matrices_by_profile` concurrently across all profile values).

Issue #1291, opened 2025-10-07
Routing phase calls OSRM sequentially instead of in parallel (if _geometry)


michael-struwe-mischok:
We found that the routing phase does OSRM calls in a sequential for loop, which means the OSRM response time + network latency is added up / multiplied times the number of routes in the result. Instead all the route requests can be done in parallel, which basically takes the maximum of these times instead of the sum.

For us this is pretty big since we are caching the OSRM table responses + supplying a [hint](https://github.com/VROOM-Project/vroom/blob/master/docs/API.md#in-solving-mode), so that the loading & solving phases are both under 100ms. The routing phase takes sth like 700ms currently & according to measurements so far, would probably improve 5x-10x by fetching in parallel.

PR #1292, opened 2025-10-07
Parallelize routing

michael-struwe-mischok:
## Issue

https://github.com/VROOM-Project/vroom/issues/1291

## Tasks

 - [ ] review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - Remove heuristic synchronisation (#1188)
 - Refactor gain evaluation in operators to ease code maintenance and extension (#1266)
 - Remove `LOG_LS_OPERATORS` (#1263)
+- Do `add_geometry` OSRM requests in parallel in routing phase (#1291)
 
 #### CI
 
--- a/src/structures/vroom/input/input.cpp
+++ b/src/structures/vroom/input/input.cpp
@@ -9,6 +9,7 @@ All rights reserved (see LICENSE).
 
 #include <algorithm>
 #include <mutex>
+#include <semaphore>
 #include <thread>
 
 #if USE_LIBOSRM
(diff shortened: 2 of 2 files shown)


jcoupey:
Thanks for submitting a PR! This fixes #1218.

jcoupey:
I've been running further tests with a remote osrm server and a low bandwidth. Already having a few threads (4 or 8) makes a huge difference: the overall routing time is more than 6x faster. Then for e.g. 32 and 64 threads, the overall routing time is slightly higher so it looks like the threads boilerplate exceeds the benefit.

Maybe that would hint toward using a max number of 32, which would also be in line with the rest of the parallelization strategy. @michael-struwe-mischok do you have other input? what do you think?

michael-struwe-mischok:
From how I understand the mentions of `LeastMaxValue` and `max` in https://en.cppreference.com/w/cpp/thread/counting_semaphore.html, we can use the template value to choose a maximum for the semaphore (just that the actual maximum delivered by the implementation might be higher if it wants to). E.g. if we say 128, then even if nb_thread is 200, the semaphore will begin its internal counter at 128 instead of 200 (or it may begin at 200, if the implementation wants to provide a higher maximum). So it's a bit like allocating enough memory for what we need.

The 128 was just a guess on how many parallel requests might be a good idea. We could also choose a higher maximum, e.g. 512. Someone might configure their OSRM to be powerful enough for this many requests / their network conditions may make it appropriate. (I assume the overhead of a too-high maximum is pretty low)

Maybe we shouldn't use `nb_thread` for the number of parallel requests & configure it separately instead. The `nb_thread` is more about how much CPU is available for VROOM, and this is more about how many requests to OSRM should be in-flight at once. The default could be sth like `nb_thread * 4`

Example scenario:
* Someone runs VROOM and OSRM on two machines that both have 4 cores
* They configure both VROOM and OSRM to use 4 threads
* Now if VROOM would send nb_thread requests in parallel, this would send a burst of 4 requests, all of them would come back at once, then send another burst of 4 requests
* Instead, if VROOM sends e.g. 32 requests in parallel, but OSRM can only actually handle 4 requests at once: It would send 32 requests, then get a burst of 4 requests back & at that moment send the next 4 requests. While the 4 responses are flying over the network, OSRM still has enough work to do with the other 28 requests to keep its CPU resources used
* If the number of parallel requests is high enough to fully use the OSRM CPU resources, but otherwise not overwhelm OSRM, that is the optimal value. Given the network as a constant that we cannot change, the bottleneck we have to exploit is that the OSRM CPU is busy with our requests. Total time = network-towards-OSRM + OSRM-CPU-for-all-the-requests + network-back-to-VROOM

However in an ideal setup, probably you should run OSRM and VROOM on the same machine to minimize the network time. In that case, `nb_thread` should already be a pretty good value.

jcoupey:
You're right that this should ideally be configured differently, but it's kind of very routing specific so I don't really feel like having another dedicated parameter for that.

As your example shows, this is highly dependent on the routing deployment (not only `osrm-routed` but potential load balancing on top of it, not to mention other routing engines) so it's hard for us to make a generic guess. What we want to avoid is having regressions where users suddenly start hitting errors because we changed the routing process internally, so we have to be careful/conservative about the defaults.

If using `std::counting_semaphore<32> semaphore(nb_thread);` as a default we should be safe, while transparently providing a speedup by default. If someone uses `-t 4`, it might indeed be less efficient on the routing side than using 32 threads as your example points out, but that's still a huge improvement over the previous behavior. 

On the other hand, users that have a routing setup allowing higher request rates can increase the default value (we could even make the `128` or `32` a `constexpr` variable configurable somewhere). @michael-struwe-mischok what do you think?

michael-struwe-mischok:
:+1: In general I don't want to block this with details so feel free to just do sth that seems appropriate :-)

I think `std::counting_semaphore<32> semaphore(nb_thread);` is fine, just two nitpicks:
* If the 32 is there as a maximum in order to avoid breakage in some situations, it would be better to do sth like `std::counting_semaphore<32> semaphore(min(32, nb_thread));` so that the maximum actually applies. Because the implementation of `counting_semaphore` could use a higher maximum than what we say in the template value
* I think we could do sth like `nb_thread * 2` here which should be a bit faster while still being careful

PR #1350, opened 2026-08-03
Fix break ordering after shipment delivery

### LINKED ISSUE #1345: Break not scheduled after a constrained delivery - shipment dropped as unassigned
When a shipment keeps the vehicle loaded right up to a delivery time window, and a break's time window opens at/after that delivery deadline, VROOM does not place the break after the drop-off. Instead it reports the shipment as unassigned, even though a valid solution exists (deliver, then take the break while empty, then return).

It looks like the break is only considered for placement in the idle time before the delivery (where the vehicle is still loaded), and is never deferred to the empty period after the delivery.

**To reproduce :** 
```json
{
  "vehicles": [
    {
      "id": 1,
      "profile": "car",
      "start_index": 0,
      "end_index": 0,
      "time_window": [28800, 57600],
      "breaks": [
        { "id": 1, "service": 1800, "time_windows": [[40800, 50400]] }
      ]
    }
  ],
  "shipments": [
    {
      "pickup":   { "id": 10, "location_index": 1, "service": 0 },
      "delivery": { "id": 11, "location_index": 2, "service": 0,
                    "time_windows": [[39900, 40800]] }
    }
  ],
  "matrices": {
    "car": {
      "durations": [
        [   0, 1800, 1800],
        [1800,    0, 1800],
        [1800, 1800,    0]
      ]
    }
  }
}
```

Here the delivery window is [39900, 40800] (11:05–11:20) and the break window is [40800, 50400] (11:20–14:00), i.e. it opens only after the delivery deadline — so the only possible placement is after the drop-off.

**Expected:** 
one route, 0 unassigned. A valid schedule exists, e.g.
start → pickup(10) → delivery(11) @ 40800 → break(1) @ 40800–42600 → end @ 44400, all within the vehicle and break windows.

**Actual: code:** 
0, summary.unassigned: 2, with pickup 10 and delivery 11 unassigned and no route using the vehicle.

**Additional notes:**
- Removing the breaks entry makes the problem solvable (the route itself is feasible).
- Widening the break window so it opens before the loaded leg (e.g. [28800, 57600]) also makes it solvable — the break is then placed before/around the delivery. The problem only appears when the break window opens during/after the loaded delivery leg, when an after-delivery placement is required.
- Adding "max_load": [0] to the break (break only when empty) does not change the outcome.

william-xue:
## Issue

Fixes #1345.

## Root cause

When evaluating a pickup while a break is pending, `TWRoute::order_choice` returned early as soon as `pickup -> break -> delivery` missed the delivery deadline. That skipped the pickup-specific fallback which also evaluates `pickup -> delivery -> break`, so a feasible shipment was rejected.

## Tasks

- [x] Continue pickup-specific ordering checks when the immediate break placement misses the next task
- [x] Add a regression instance for the reported delivery and break windows
- [x] Verify the route has no unassigned tasks and orders pickup, delivery, then break

## Validation

- `make -j2 USE_ROUTING=false`
- Regression before fix: 0 routes, 2 unassigned tasks
- Regression after fix: 1 route, 0 unassigned tasks; delivery arrival 39900, break arrival 40800, route end 43500
- Existing `docs/example_2.json` output matches `docs/example_2_sol.json`
- `git diff --check`

### DIFF
--- a/.github/workflows/vroom.yml
+++ b/.github/workflows/vroom.yml
@@ -34,6 +34,13 @@ jobs:
         working-directory: src
       - name: Validate output
         run: diff <(bin/vroom -i docs/example_2.json  | jq '.routes[].steps[]' --sort-keys) <(jq '.routes[].steps[]' --sort-keys docs/example_2_sol.json)
+      - name: Validate pickup-delivery break ordering
+        run: |
+          bin/vroom -i tests/issue_1345.json | jq -e '
+            .summary.unassigned == 0 and
+            [.routes[0].steps[].type] == ["start", "pickup", "delivery", "break", "end"] and
+            .routes[0].steps[2].arrival <= 40800 and
+            .routes[0].steps[3].arrival >= 40800'
       - name: Build libvroom example
         run: make -j
         env:
--- a/src/structures/vroom/tw_route.cpp
+++ b/src/structures/vroom/tw_route.cpp
@@ -546,8 +546,10 @@ OrderChoice TWRoute::order_choice(const Input& input,
     job_then_break_end = earliest_job_end + b.service;
   }
 
-  if (job_then_break_end + travel_after_break > next.latest) {
-    // Starting the break is possible but then next step is not.
+  if (job_then_break_end + travel_after_break > next.latest &&
+      j.type != JOB_TYPE::PICKUP) {
+    // Starting the break is possible but then next step is not. For a
+    // pickup, keep checking because pickup -> delivery -> break may fit.
     oc.add_break_first = true;
     return oc;
   }
--- a/tests/issue_1345.json
+++ b/tests/issue_1345.json
@@ -1,0 +1,34 @@
+{
+  "vehicles": [
+    {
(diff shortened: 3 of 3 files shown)


william-xue:
Closing this for now — not because the fix is invalid, but because there is no review signal on it.

Summary of what is on the branch: it restores feasible pickup → delivery → break ordering in a one-vehicle PDPTW when the delivery and break windows are disjoint. Regression on the bundled instance: before 0 routes / 2 unassigned, after 1 route / 0 unassigned, objective cost/duration 5,400, with precedence, time-window and vehicle-horizon invariants satisfied. Both Sonar checks pass; the Actions workflow stays in `action_required` because it needs fork-workflow approval upstream.

The branch remains at william-xue:vroom (fix/break-ordering-shipment) with the regression test, so it is a one-command revive. I am happy to reopen, rebase and adjust to whatever direction you prefer as soon as there is interest — I just do not want to keep it sitting in the queue looking like an active proposal.

